Issue #106, opened 2019-06-18
Avoid writing (then reading) JSON form at the start of `SDA_query`


bocinsky:
Currently, the `SDA_query` function writes a JSON form to a temporary file, then reads the form as part of the `httr::POST` request. Instead, we can just define the query within the request:
```{r}
  # submit request
  r <- httr::POST(url = "https://sdmdataaccess.sc.egov.usda.gov/tabular/post.rest",
                  body = list(query = q,
                              format = "JSON+COLUMNNAME"),
                  encode = "form")
```

dylanbeaudette:
Good idea. Not sure why we wrote to a file first. This is much simpler. Tested and appears to work well.

PR #107, opened 2019-06-18
Avoid i/o of JSON form at the start of `SDA_query`

bocinsky:
Implements change suggested in https://github.com/ncss-tech/soilDB/issues/106


dylanbeaudette:
Tested with soilDB test suite and appears to work as expected. Thanks @bocinsky!

Issue #108, opened 2019-06-18
Avoid i/o in `.post_process_SDA_result_set` when post-processing `SDA_query` data


bocinsky:
For large queries, the strategy of writing the table and then reading it using `utils::read.table` causes significant overhead. Try and keep everything in memory.

dylanbeaudette:
Tested with soilDB test suite and appears to work as expected. Thanks @bocinsky!

PR #109, opened 2019-06-18
Avoid i/o in `.post_process_SDA_result_set`

bocinsky:
Addresses https://github.com/ncss-tech/soilDB/issues/108


bocinsky:
Should be tested further. What are the "go to" queries for testing results from SDA?

dylanbeaudette:
Tested with soilDB test suite and appears to work as expected. Thanks @bocinsky!

Issue #38, opened 2017-12-04
SDA and duplicate records


dylanbeaudette:
Trying to get my head around this one. 

An example from Stephen:
```r
library(soilDB)
library(daff)

# test both cases, thanks Stephen for the example
x <- get_component_from_SDA(WHERE = "muname = 'Millsdale silty clay loam, 0 to 2 percent slopes'", duplicates = TRUE)
y <- get_component_from_SDA(WHERE = "muname = 'Millsdale silty clay loam, 0 to 2 percent slopes'", duplicates = FALSE)

# get differences
d <- diff_data(x, y)
render_diff(d, title='duplicates=TRUE vs. duplicates=FALSE')
```

The two calls to `get_component_from_SDA()` generate the following SQL:

```sql
-- duplicates == TRUE
SELECT mukey, mu.nationalmusym, compname, comppct_r, compkind, majcompflag, localphase, slope_r, tfact, wei, weg, drainagecl, elev_r, aspectrep, map_r, airtempa_r, reannualprecip_r, ffd_r, nirrcapcl, nirrcapscl, irrcapcl, irrcapscl, frostact, hydgrp, corcon, corsteel, taxclname, taxorder, taxsuborder, taxgrtgroup, taxsubgrp, taxpartsize, taxpartsizemod, taxceactcl, taxreaction, taxtempcl, taxmoistscl, taxtempregime, soiltaxedition, cokey 
FROM legend l 
INNER JOIN
mapunit mu ON mu.lkey = l.lkey 
INNER JOIN (
SELECT compname, comppct_r, compkind, majcompflag, localphase, slope_r, tfact, wei, weg, drainagecl, elev_r, aspectrep, map_r, airtempa_r, reannualprecip_r, ffd_r, nirrcapcl, nirrcapscl, irrcapcl, irrcapscl, frostact, hydgrp, corcon, corsteel, taxclname, taxorder, taxsuborder, taxgrtgroup, taxsubgrp, taxpartsize, taxpartsizemod, taxceactcl, taxreaction, taxtempcl, taxmoistscl, taxtempregime, soiltaxedition, cokey , mukey AS mukey2 FROM component
) AS c ON c.mukey2 = mu.mukey 
WHERE muname = 'Millsdale silty clay loam, 0 to 2 percent slopes' 
ORDER BY cokey, compname, comppct_r DESC;
```

```sql
-- duplicates == FALSE
[...]

PR #121, opened 2020-01-22
demoting fetchNASIS_pedons() and fetchNASIS_components() to internal functions

smroecker:
As discussed I demoted fetchNASIS_pedons() and fetchNASIS_components to internal functions. They will no longer show up in the list of soilDB functions or help files.

### DIFF
--- a/R/fetchNASIS.R
+++ b/R/fetchNASIS.R
@@ -1,7 +1,16 @@
 # convenient interface to local NASIS data
 # from: pedons | components | lab | ???
 # ... : arguments passed on to helper functions
-fetchNASIS <- function(from='pedons', url = NULL, ...) {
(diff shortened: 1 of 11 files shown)


brownag:
RE: #38 - SDA and duplicate records b1e713a

New warning from `get_component_from_SDA` when `duplicates=FALSE` (default) and `mukey` is in WHERE clause
```
 if(!duplicates & grepl(WHERE, pattern = "mukey")[1])
    warning("duplicates is set to FALSE and 'mukey' is in WHERE clause. Note: 'mukey' omitted from result.", call.=FALSE)
```

New entry in doc for `duplicates` argument to `fetchSDA_component`/`get_component_from_SDA`

> logical; if TRUE a record is returned for each unique mukey (may be many per nationalmusym)

Issue #141, opened 2020-09-03
Enhance `SDA_spatialQuery(what="geom")` to query from either SSURGO or (new) STATSGO?


dschlaep:
The function `SDA_spatialQuery` (soilDB v2.5.7) currently returns geometries and `mukey` values from `SSURGO` and not from `STATSGO`. As long as `SSURGO` remains incomplete and `SDA` doesn't connect to merged version such as `gNATSGO`, it would be great to have the functionality to query from either database for a complete workflow coverage.

A call with the argument `what = "mukey"` queries the tabular table `mapunit` and relies on a SQL statement based on `SDA_Get_Mukey_from_intersection_with_WktWgs84()` to handle the spatial conversion; this, however, restricts the output to `SSURGO`. Until there is a comparable function for `STATSGO` not much can be done for this case apparently.

However, a call with the argument `what = "geom"` relies on T-SQL intrinsic spatial function `STIntersects` that queries the spatial table `mupolygon` of `SSURGO`. According to a comment from Dec 2017 made in response to an earlier, different issue, a corresponding table exists for `STATSGO` https://github.com/ncss-tech/soilDB/issues/36#issuecomment-349054067, i.e., `gsmmupolygon`.

I created a local copy of `SDA_spatialQuery` for my own use with a new argument `db = c("SSURGO", "STATSGO")` and few minor code changes, which appear to work well (see below).

As a simple user, I don't know whether querying `gsmmupolygon` is a supported activity. If it is, however, then it would be great if you would consider whether the suggested enhancement could be a good fit for `soilDB`! Let me know if you would like me to create a pull-request with the suggestion incorporated.

Thanks!

```
SDA_spatialQuery2 <- function(geom, what='mukey', db = c("SSURGO", "STATSGO"), geomIntersection=FALSE) {
  ...

  db <- match.arg(db)

  if (what == "mukey" && db == "STATSGO") {
    stop("query type 'mukey' for 'STATSGO' is not supported", call. = FALSE)
  }
  
  ...

  if(what == 'geom') {
    # as suggested in https://github.com/ncss-tech/soilDB/issues/36#issuecomment-349054067
    db_table <- switch(db, SSURGO = "mupolygon", STATSGO = "gsmmupolygon")

    # return intersection
    if(geomIntersection) {
      q <- sprintf("
               SELECT
                 mupolygongeo.STIntersection( geometry::STGeomFromText('%s', 4326) ).STAsText() AS geom, P.mukey
[...]

PR #142, opened 2020-09-03
Function `SDA_spatialQuery()` gains new argument `db`

dschlaep:
- addressing issue #141

- function `SDA_spatialQuery()` gains new argument `db` to select whether geometries should be queried from Soil Geographic Database SSURGO or STATSGO
- default value of `db` is "SSURGO"; the new argument is backwards compatible
- updated roxygen-style function documentation in source file and in man/*.Rd object
- updated unit tests to include a simple check for the new argument with default ("SSURGO") and new "STATSGO" values

### DIFF
--- a/R/SDA-spatial.R
+++ b/R/SDA-spatial.R
@@ -75,7 +75,10 @@ processSDA_WKT <- function(d, g='geom', p4s='+proj=longlat +datum=WGS84') {
 #' @param geom a Spatial* object, with valid CRS. May contain multiple features.
 #' @param what a character vector specifting what to return. `mukey`: `data.frame` with intersecting map unit keys and names, `geom` overlapping or intersecting map unit polygons
 #' @param geomIntersection logical; FALSE: overlapping map unit polygons returned, TRUE: intersection of `geom` + map unit polygons is returned.
-#' 
+#' @param db a character vector identifying the Soil Geographic Databases
+#'   (`SSURGO` or `STATSGO`) to query. Option \var{STATSGO} currently works
+#'   only in combination with \code{what = "geom"}.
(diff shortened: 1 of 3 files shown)


brownag:
I am very excited that this functionality has been implemented. I recently needed something like this -- and I didn't ultimately implement it via SDA. 

I pulled your changes, tested new argument, and ran tests+checks. 


Quick little demo of fetching STATSGO geom for a SSURGO survey area boundary:
```
library(soilDB)
library(sf)

# arbitrary extent; in this case a SSA boundary from a Mariposa County (CA649)
ext <- read_sf("FGCA649_Projects_2019_1125_ra.gdb", "ca649_b")

# transform to GCS
ext <- st_transform(ext, st_crs("+proj=longlat +datum=WGS84"))

# do spatial query against STATSGO, returning "geom"
foo <- SDA_spatialQuery(as_Spatial(ext), what = "geom", db = "STATSGO")

# inspect result
plot(ext$Shape, lwd = 2, col = "RED")
plot(foo, add = TRUE)
```

image

Good work @dschlaep!

I would personally love to see the STATSGO-equivalent of `SDA_Get_Mukey_from_intersection_with_WktWgs84` but as you have assessed in #141 we may be stuck until such a feature is added to SDA. @dylanbeaudette thoughts on this?

Overall, these changes look good to me.

dylanbeaudette:
Excellent, thank you @dschlaep for the idea, and @brownag for the example--you saved me 20 minutes of work this morning. I'll probably go in an make some minor adjustments next.

dschlaep:
Many thanks for taking this up so quickly!

Issue #146, opened 2020-09-29
switch from RODBC to DBI/odbc for connections to the local NASIS database


dylanbeaudette:
Tiny example.
```r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), ...)
```

This started as #145.

brownag:
Reprex for dealing with issues related to long (text) columns, truncation to 255 characters, and possible stumbling blocks in conversion to DBI. I thought that with the most recent {odbc} that `dbGetQuery` would not error, regardless of long ccolumn order, but apparently not.

``` r
# reprex for issues with column ordering and ODBC drivers

# RODBC truncates long columns to 255 characters
library(RODBC)
con2 <- soilDB:::.openNASISchannel()

res1 <- sqlQuery(con2, "SELECT * FROM calculation")
max(nchar(res1$calc)) # truncated to 255 characters
#> [1] 255
# manual casting of columns to ntext works, but not for simple queries like SELECT *

# compare DBI
library(DBI)
library(odbc) # driver to read MSSQL 
#> Warning: package 'odbc' was built under R version 4.0.3

con <- dbConnect(odbc::odbc(), 
                 DSN = "nasis_local", 
                 UID = "NasisSqlRO",
                 PWD = "nasisRe@d0n1y365")

# this produces an error; may be fixed with new odbc? 
#  see https://github.com/r-dbi/odbc/issues/309
dbGetQuery(con, "SELECT * FROM calculation")
#> Error in result_fetch(res@ptr, n): nanodbc/nanodbc.cpp:3011: 07009: [Microsoft][SQL Server Native Client 11.0]Invalid Descriptor Index
#> Warning in dbClearResult(rs): Result already cleared

# this produces same error
library(dplyr, warn.conflicts = FALSE)
library(dbplyr)
#> 
#> Attaching package: 'dbplyr'
#> The following objects are masked from 'package:dplyr':
#> 
#>     ident, sql
tbl(con, "calculation")
#> Error in result_fetch(res@ptr, n): nanodbc/nanodbc.cpp:3011: 07009: [Microsoft][SQL Server Native Client 11.0]Invalid Descriptor Index
#> Warning in dbClearResult(res): Result already cleared

# this works without error (manually identify and reorder long columns to end)
libary(tidyselect)
#> Error in libary(tidyselect): could not find function "libary"
long_cols <- odbcConnectionColumns(con, "calculation") %>%
  filter(field.type == "varchar" & column_size == 0) %>%
[...]

Issue #145, opened 2020-09-28
text fields in local NASIS are truncated at 255 characters


dylanbeaudette:
This seems to have occured after the recent switch to a newer ODBC driver.

Put something into your selected set and try it.
```r
library(RODBC)

q <- "SELECT projectiid, uprojectid, projectname, projectdesc
  
  FROM 
  project_View_1
  ;
  "
# setup connection local NASIS
channel <- RODBC::odbcDriverConnect(connection=getOption('soilDB.NASIS.credentials'))

# exec query
d <- RODBC::sqlQuery(channel, q, stringsAsFactors=FALSE)

# close connection
RODBC::odbcClose(channel)

d
```

brownag:
Interesting. Perhaps an interaction with {RODBC} package, as DBI/odbc I get max 1500 characters not 255.

```
library(RODBC)

q <- "SELECT projectiid, uprojectid, projectname, projectdesc
  
  FROM 
  project_View_1
  ;
  "
# setup connection local NASIS
channel <- soilDB:::.openNASISchannel()

# exec query
d <- RODBC::sqlQuery(channel, q, stringsAsFactors=FALSE)

# close connection
RODBC::odbcClose(channel)

max(nchar(d$projectdesc), na.rm = T)
# 255
```

```
library(DBI)
con <- dbConnect(odbc::odbc(), 
                 DSN = "nasis_local", 
                 UID="NasisSqlRO",
                 PWD="nasisRe@d0n1y365")
res <- dbSendQuery(con, q)
dat <- dbFetch(res)
max(nchar(dat$projectdesc), na.rm=T)
# 1500
```

dylanbeaudette:
That is odd, RODBC has worked for a long time without issue. Any thoughts on switching over? Is the 1500 limit a self-imposed limit or the driver?

brownag:
I have wondered whether something might happen that would tip the tables in favor of DBI/odbc. Until that little snafu with R 4.0/Rcpp i hadn't seriously considered it.

If i drop the _View_1 from the query I get 9999 which IIRC is the max for that field in nasis. So, 1500 is max in my selected set (I have ~all 2-SON projects I think)

brownag:
Castting varchar to ntext seems to do the trick with {RODBC}.

```
library(RODBC)

q <- "SELECT projectiid, uprojectid, projectname, CAST(projectdesc AS ntext) as 'projectdesc'
  
  FROM 
  project_View_1
  ;
  "
[...]

PR #149, opened 2020-12-29
Convert NASIS database access from RODBC to DBI/odbc

brownag:
This Pull Request converts all usages of {RODBC} for accessing the NASIS local database to use {DBI}/{odbc}.

There are several cases where the ODBC 11 or 13 drivers available to us on CCE machines may not work correctly for generic queries written against the NASIS database, but all functions and queries built into {soilDB} return correct values.

## Why {DBI}?

Notably the conversion to {DBI} allows us to drop in other database drivers, with only minor changes to existing queries. This allows us to utilize the NASIS schema in other "containers." The currently available tools are basic e.g. create a database from your selected set via `createStaticNASIS`. This uses the `RSQLite` driver to create file-based SQLite databases. All of the standard NASIS access methods work against those databases, as well as generic queries--which are generally better behaved via the RSQLite driver than they are with ODBC 13 driver + {odbc}. Only minor syntax changes (notably removing RIGHT JOIN and T-SQL specific syntax) were required to support SQLite, and it is possible that other drivers can be implemented.

A static database may be quite convenient for managing "multiple selected sets" that may be project specific and generally facilitates construction of self-contained databases using the NASIS data model.

---

RE: #145 #146

- [x] Check if any commits can/need to be cherry-picked off nasisLITE branch. 
- results from old == new?
   - [x] storage of `master` branch results of running [defined] methods (save `test` object from <strike>[misc/validate-NASIS-results.R](https://github.com/ncss-tech/soilDB/blob/370ea0b02792db347a348ea45b354d55277bfabe/misc/validate-NASIS-results.R)</strike>https://github.com/ncss-tech/soilDB/blob/nasisDBI/misc/run-all-NASIS-get-methods.R, then checkout nasisDBI branch and compare)
   - [x] GitHub Label for PRs that impact NASIS tests
   - [x] Move VARCHAR(MAX) fields to end of query, per MSSQL specs (1f872e9)
   - [x] tabulate results (we know column names changed order per above)
     - [x] resolve issue with overflow of SQL integer columns (e.g. site aspect) for missing values
   
- syntax for `dbConnectNASIS` and `dbQueryNASIS`
   - [x] function names and aliases (added `NASIS()` alias for `dbConnectNASIS`)
   - [x] vectorization over multiple queries (`dbQueryNASIS` now supports named vector of queries for `q`, returning list result)

- [x] investigate stability of open-ended queries e.g. `SELECT * FROM foo` -- are invalid descriptor errors a side effect of our older MSSQL driver?
  - NOTE: see Discussion #166 for info on the best available ODBC drivers that are more compatible with _nanodbc_


dylanbeaudette:
Very cool, thanks for doing all of this work. I'm sure you have tried this, but I have to ask: results from old == new?

brownag:
This is a draft PR for commits from 2 branches that hit many functions so I have not specifically tested for equivalence e.g. with {daff} of objects produced by the different branches. That certainly could be done before this is marked ready for review, but I was not planning on doing that this week. That idea fits well with the needs we have for improving testing of NASIS functionality. 

Stephen asked me if/when I was going to merge this so I created the PR to inspect and harmonize the changes that cannot be merged from nasisLITE branch.

Please clone and run tests on your local if you are curious. I have actually not done that -- for this PR -- on my government machine. The remote tests pass. Results *should* be equivalent. I don't think anything was changed that would  affect object contents. But as with all NASIS things testing is nuanced and sort of depends on your selected set.

dylanbeaudette:
I can help with testing by creating some before / after datasets. Changes appear superficial, but we are switching to an entirely new DB abstraction layer. I'm particularly interested in how long text fields will be handled--most of the interpretation engine relies on extracting XML from "long" text fields.

brownag:
Correct me if I am wrong -- but the only methods from soilDB currently used by the interpretation engine are `SDA_query` and `uncode`. Is there a new version?

I certainly will not close or merge this without verifying that all of the things identified previously in #146 are covered.

dylanbeaudette:
The interpretation engine code (https://github.com/ncss-tech/interpretation-engine) is in flux, but the extraction of rules, evaluations, and properties requires interacting with the local NASIS client. That said, I forgot that it uses its own ODBC connection and not the soilDB-specific connection details. So, I'll manually move this over to that code base after this PR is complete. In short, false alarm.

brownag:
Thanks @dylanbeaudette for testing and the comments. 

Another thing I would like you both to consider here is syntax for the new methods `dbConnectNASIS` and `dbQueryNASIS`.

These methods are optimized for "one-liners" against the local NASIS db. I considered an even shorter alias for `dbConnectNASIS` that was just `NASIS()`

```
dbQueryNASIS(dbConnectNASIS(), "SELECT * from foo")
```

By default, `dbQueryNASIS` closes the connection after the query. I observed that the vast majority of our connections were used for sending just a single query and that many people, myself included, forget to close connections.

To keep the connection open, you set `close = FALSE`. This is used in e.g. `get_extended_data_from_NASIS_db` where a sequence of queries can be executed using same connection. Open to feedback on function/argument names.

brownag:
[nasisDBI](https://github.com/ncss-tech/soilDB/tree/nasisDBI) branch is passing local tests and R CMD check w/ selected set containing diverse pedon/vegplot data. Local NASIS ODBC connection for relevant `fetch`/`get` methods via DBI.

[misc/validate-NASIS-results.R](https://github.com/ncss-tech/soilDB/blob/370ea0b02792db347a348ea45b354d55277bfabe/misc/validate-NASIS-results.R) shows semi-automated checking of exported methods by file name / method name. I used it to fix the above things (https://github.com/ncss-tech/soilDB/pull/149/commits/1f872e931543857502e08c2e60f6e82015bb43f0, https://github.com/ncss-tech/soilDB/pull/149/commits/e655a7d12e4ba63b5bf3b31c77bca5e5e5e30387) and create some new unit tests (https://github.com/ncss-tech/soilDB/pull/149/commits/c1506878dfe40bd96e73ec7870c346f19e235bcd)

dylanbeaudette:
I plan on old / new testing tomorrow AM with a large pile of components / pedons. The existing tests look great.

brownag:
Great!  I reorganized the checklist in original post a little bit.

I am going to move two items to a new issue, as more generalized workflows for testing and validating NASIS functionality will follow after this PR is merged and a few more puzzle pieces are put into place.

  - _.Rda_ stored as artifact of `R CMD check` Action? 
  - Presence of NASIS-local tag triggers comparison against `master`

brownag:
With most recent merged PR, this type of soilDB syntax is now possible.

``` r
library(soilDB)

nasislite_path <- "C:/Geodata/soils/NASIS-data.sqlite"

res <- dbQueryNASIS(dbConnectNASIS(nasislite_path), "SELECT geomfname, geomfiid, geomftiidref FROM geomorfeat")

head(res)
#>       geomfname geomfiid geomftiidref
#> 1          alas      122            1
#> 2  alluvial fan        2            1
#> 3 alluvial flat        3            1
#> 4     anticline      123            1
#> 5         arete        4            1
#> 6        arroyo        6            1
```

Next is to refine the "dump" method that writes the relevant tables from selected set / local NASIS to .sqlite file. 

Above works with the pedon snapshot that is floating around from 2018 for one-off queries -- which was the impetus for me developing all of this. Note there are several issues with it compared to the official NASIS schema -- truncation of column names to X characters, uppercase/lowercase/underscores, no need for metadata uncoding etc. 

The nasisLITE branch "fixed" or at least patched those things to make it work, but this branch will ultimately contain a more general solution that works seamlessly with the standard NASIS data model.

brownag:
Or, perhaps one is into {dplyr}/{dbplyr} style in their scripts...

``` r

library(soilDB)
library(dplyr, warn.conflicts = FALSE)
#> Warning: package 'dplyr' was built under R version 4.0.3

nasislite_path <- "C:/Geodata/soils/NASIS-data.sqlite"
con <- dbConnectNASIS(nasislite_path) 

con %>%
  tbl("geomorfeat") %>% 
  select(everything()) %>%
  collect()
#> # A tibble: 732 x 3
#>    geomfname        geomfiid geomftiidref
#>    <chr>               <int>        <int>
#>  1 alas                  122            1
#>  2 alluvial fan            2            1
#>  3 alluvial flat           3            1
#>  4 anticline             123            1
#>  5 arete                   4            1
#>  6 arroyo                  6            1
#>  7 ash flow              124            1
#>  8 atoll                 125            1
#>  9 avalanche chute       126            1
#> 10 avalanche debris      147            1
#> # ... with 722 more rows
```

<sup>Created on 2021-01-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

brownag:
RE: #165 -- it appears SQL Server 13 driver fixes the invalid descriptor index errors with `SELECT *` but introduces a "new" data type conversion problem.

COMPARE: `dbQueryNASIS(NASIS(), "SELECT aspect FROM site_View_1")$aspect`
WITH: `dbQueryNASIS(NASIS(), "SELECT * FROM site_View_1")$aspect`

```r
> expect_equal(is.na(dbQueryNASIS(NASIS(), "SELECT aspect FROM site_View_1")$aspect),
+             
+              is.na(dbQueryNASIS(NASIS(), "SELECT * FROM site_View_1")$aspect))
Error: is.na(dbQueryNASIS(NASIS(), "SELECT aspect FROM site_View_1")$aspect) not equal to is.na(dbQueryNASIS(NASIS(), "SELECT * FROM site_View_1")$aspect).
9 element mismatches
```

 - with Microsoft driver version 11 (first works as expected, second results in invalid descriptor index error)
 - with Microsoft driver version 13 (first works as expected, second does not error but returns _garbage_ integer overflow values for the 9 `NA` records

Both of the above queries work with {RODBC}-based code, which has some options affecting type conversion.  Note that it also works as expected when one does `soilDB:::.dump_NASIS_table("site")$aspect` which automatically makes an "explicit" query for all columns in a table; getting around the version 11 errors and the weird version 13 behavior.

With RSQLite the nanodbc driver is not used, but the .sqlite contents would depend on what driver/method one used to make their static database in the first place. Using the internal {soilDB} dump method appears to be "safe"

Leads me to think it is either a NASIS-specific (e.g. type definitions/ranges/storage of NULL and/or poorly designed tables)  or older driver-specific.

brownag:
See https://github.com/ncss-tech/soilDB/blob/nasisDBI/misc/run-all-NASIS-get-methods.R for latest testing suite for NASIS "get" and "fetch" methods

dylanbeaudette:
Testing a SS based on all of the non-additional MU, representative DMU, component pedons + site records. Noticed:

```
> cm <- get_comonth_from_NASIS_db(fill = TRUE)
Error in new_result(connection@ptr, statement, immediate) : 
  external pointer is not valid
```

Also noticed that pedon records have some coded (?) values:
image

brownag:
> get_comonth_from_NASIS_db

Thanks! I didn't test with the `fill=TRUE` argument! The connection needs to be left open to run the second query that the `fill` arg uses.

As far as the coding/decoding from NASIS domains... I thought that was working. Will have to look closer

dylanbeaudette:
Thanks. Will test again tonight and report back.

brownag:
Great, thanks. When I saw what was going on had to try and fix.

I was wrong-- the necessary stuff to trigger correct "uncoding" wasn't uniformly in place. Further, in the absence of a NASIS database (i.e. on my home machine where I did the above commits) the `uncode()` routine would fail to get the right data for `db="NASIS"` domains. Now it should properly pull from the `static_path` if specified. The horizon data uncoding was just a mistake on my part, using wrong variable name (a copy that was never used).

dylanbeaudette (approved):
Fully tested using the CA792 legend: 
  * MU/DMU/components/component month
  * Site/Pedon

Tested:
  * before / after with {daff}
  * re-made DMU summary reports and side-by-side comparisons

Thanks!

brownag:
Note that {DBI} is in Imports and the driver packages {odbc} and {RSQLite} are in suggests.

Updated the following resources with code to get DBI/odbc/RSQLite dependencies. 

Note that since these are not in Imports/Suggests for the current version of soilDB on CRAN our default heuristic of having you install the CRAN package first to get dependencies before installing the GH version will not "work." 

Users will need to manually install the above 3 packages or use `remotes::install_github('ncss-tech/soilDB', dependencies=TRUE)` to get dependencies from GitHub `DESCRIPTION` file. As usual, do the latter in a fresh session, without any packages loaded.

The following resources have been updated to ensure the packages will be available:
 - http://ncss-tech.github.io/AQP/soilDB/setup_local_nasis.html (ODBC setup instructions)
 - http://ncss-tech.github.io/stats_for_soil_survey/chapters/0_pre-class-assignment/pre-class-assignment.html (S4SS precourse old link)
 - http://ncss-tech.github.io/stats_for_soil_survey/book/index.html (S4SS precourse new link)

PR #153, opened 2021-01-06
NASIS DBI interface to SQLite static (local file) database

brownag:
This PR implements the argument for a local SQLite file from `fetchNASIS` down through the lower level methods, on to the `dbConnectNASIS()` calls. 

This is cherry-picked so none of the special modifications made for using the (slightly different from NASIS internal) NASIS pedon snapshot (RE: nasisLITE branch) are included. All local NASIS tests and validations (see script in /misc/  made for nasisDBI) are passing.


brownag (commented):
Scanned over; merging into nasisDBI now

PR #156, opened 2021-01-12
minor enhancements for WCS SSURGO MUKEY interface

brownag:
These commits are grouped so @dylanbeaudette can take 'em or leave em.

1. proj4 safe examples -- not controversial
2. reverse order of db and aoi arguments -- does not affect example, I think aoi in first position is more pipe friendly and is just my preference
3. add sf support; this passes R CMD CHECK -- we already have sf in suggests; appears to be equivalent to list-based
4. very small extent test; downloads something like 500 bytes of data from WCS

### DIFF
--- a/R/WCS-utils.R
+++ b/R/WCS-utils.R
@@ -1,53 +1,66 @@
 
 
-# obj: list(aoi, crs) or Spatial* object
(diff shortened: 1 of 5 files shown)


dylanbeaudette:
Great ideas, thanks!

PR #163, opened 2021-01-29
Rd2roxygen, and various nasisDBI documentation improvements

brownag:


dylanbeaudette:
This comment may be too late, but per #164, it looks like a lot of strange conversions (.Rd -> roxygen) may have happened in this PR. ?

PR #176, opened 2021-03-30
Use aqp::checkHzDepthLogic for horizon depth logic checks

brownag:


R/simplifyArtifactData.R
```diff
@@ -6,75 +6,75 @@
...
-  
+
+
   ## split flat/nonflat
   idx <- grep('^flat', x$huartshp, ignore.case = TRUE, invert=TRUE)
   arts <- x[idx, ]
-  
+
   idx <- grep('^flat', x$huartshp, ignore.case = TRUE)
   farts <- x[idx, ]
```
dylanbeaudette, line 31: busted

dylanbeaudette (approved):
Thanks for doing this. Do you want have or want before / after sample data set for comparisons? I figured that this is part of your process but happy to send something over.

R/simplifyArtifactData.R
brownag: ?

brownag:
I am in the process of testing the more unusual methods. All NASIS things work without error, per the routine I have for testing all the different query methods. It is not convenient for me to work with your before/after datasets, but feel free to let me know if any of the following methods are not working as expected:

brownag:
Tested and seems to be working as expected for `fetchNASIS('pedon_report')` and `fetchPedonPC`/`getHzErrorsPedonPC` PedonPC .accdb containing new CA750 pedons from @jskovlin

R/simplifyArtifactData.R
dylanbeaudette: a joke, the variable name caught my eye and my inner 8 year old liked it.

dylanbeaudette:
> I am in the process of testing the more unusual methods. All NASIS things work without error, per the routine I have for testing all the different query methods. It is not convenient for me to work with your before/after datasets, but feel free to let me know if any of the following methods are not working as expected: fetchNASIS_report, fetchPedonPC, getHzErrorsPedonPC

I have no good ways to test those three functions, sorry.

Is the before / after testing part of this process? If not then I am offering some assistance. If it is, then feel free to ignore.

PR #180, opened 2021-04-08
adding interface to formative element grids

dylanbeaudette:


dylanbeaudette:
Formative element grids are created from the formative element dictionaries in {SoilTaxonomy}. Like this:

```r
library(SoilTaxonomy)
data('ST_formative_elements')

## subgroup
sg <- ST_formative_elements$subgroup

# remove missing connotation
sg <- sg[which(sg$connotation != ''), ]

# remove 'typic'
sg <- sg[which(sg$element != 'typic'), ]

x <- split(sg, sg$connotation)

sg.rules <- lapply(x, function(i) {
  e <- sprintf("LOWER(taxsubgrp) ~ '^%s'", i$element)
  exp <- paste(e, collapse = ' OR ')
  
  # use the first element name, avoiding duplication
  res <- data.frame(
    id = i$element[1],
    exp = exp,
    elements = paste(i$element, collapse = '/'),
    connotation = i$connotation[1],
    stringsAsFactors = FALSE
  )
  
  return(res)
})

sg.rules <- do.call('rbind', sg.rules)
sg.rules <- sg.rules[order(sg.rules$id), ]

# save to CSV for iteration over via GNU parallel / psql client

# make associated manual pages
cat(paste0(sprintf("  * %s: (%s) %s", sg.rules$id, sg.rules$elements, sg.rules$connotation), collapse = '\n'))
```

dylanbeaudette:
> Yep, I commented on it before I properly realized that it is more of a SoilTaxonomy dataset issue. I think what you did above was fine in that it follows the correct pattern of using the last occurrence as exists in current LUT.
> 
> I guess I am just wondering if it just be that the second instance is "thapt" not "thapto"? We might be able to remove most of these and just use the roots. I didn't really re-define the lookup table at all when I revamped the way `FormativeElements` worked, perhaps the definitions could just be simpler overall.

I'd like to refine the formative elements dictionaries and methods in SoilTaxonomy after this PR is merged. We can clarify the 'thatp' issue when adding all of the missing SG formative elements (https://github.com/ncss-tech/SoilTaxonomy/issues/5). Long-term the creation / documentation of these grids will be based on a more specific organization of the concepts, keyed to unique formative element IDs or "roots".

brownag (approved):
I think these grid names should be good to go

PR #183, opened 2021-05-14
CRAN Release 2.6.2

brownag:
It appears that a couple tests of the WCS functionality are erroring out on new M1 Mac platform. Need to fail gracefully. 

https://www.stats.ox.ac.uk/pub/bdr/M1mac/soilDB.out
 - I have replaced a hard `stop()` on "bad WCS request" in `ISSR800.wcs` and `mukey.wcs` with a message and a try-error result.  

- The documentation has been updated to reflect that a `try-error` will be returned when WCS fails. 
- Tests are skipped if offline and on CRAN
- Test expectations have been updated to handle possible outputs under a failed WCS request (if for some reason tests fail to be skipped)

I will be making a submission to CRAN by end of today to correct it.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,6 @@ Package: soilDB
 Type: Package
 Title: Soil Database Interface
 Version: 2.6.2
-Date: 2021-04-22
 Authors@R: c(person(given="Dylan", family="Beaudette", role = c("aut"), email = "dylan.beaudette@usda.gov"),
              person(given="Jay", family="Skovlin", role = c("aut")), 
              person(given="Stephen", family="Roecker", role = c("aut")), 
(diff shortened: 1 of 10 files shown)


brownag:
Passing R CMD check `rhub::check_for_cran()` for 
- Windows Server 2008 R2 SP1, R-devel, 32/64 bit,
- Ubuntu Linux 20.04.1 LTS, R-release, GCC
- Fedora Linux, R-devel, clang, gfortran

dylanbeaudette:
Thanks for these changes. I'll have to ask you sometime about `rhub`, looks more convenient than win-builder.

brownag:
> I'll have to ask you sometime about `rhub`, looks more convenient than win-builder.

rhub is handy and good for testing some of those lesser-used/harder-to-set-up platforms. I am not sure I totally trust it as a  replacement for `devtools::check_win_devel()`, but choosing the setups on rhub that do the CRAN checks is probably almost as good and at least minimally satisfies checking `--as-cran` against R-devel. 

IMO rhub is especially useful when win-builder is taking >>>30mins to do a check and I want to verify something is fixed. Also I like to run `rhub::check_for_cran()` in a script via RStudio _Tools_ >> _Jobs_ >> _Start Local Job..._ so it runs in background.

Issue #55, opened 2018-02-13
profiling: fetchNASIS_pedons()


dylanbeaudette:
Adding to misc/ dir of package. Initial results can be seen [here](http://ncss-tech.github.io/soilDB/misc/fetchNASIS_pedons-flame-graph.svg).

It looks like `simplifyFragmentData` could use some work.

brownag:
I like these flame graphs. 

I was surprised to see simplifyFragmentData so prevalent here (~42.5%)... always figured it was the simplifyColorData (only 12.6%) that made this function hang up with bigger datasets.

dylanbeaudette:
After the [latest fixes](https://github.com/ncss-tech/soilDB/commit/8e6b0d389df5302d4c8823c7281c18d511628061) applied to `simplifyFragmentData`, this function now contributes much less to the overall time.

   * http://ncss-tech.github.io/soilDB/misc/fetchNASIS_pedons-flame-graph.svg
   * http://ncss-tech.github.io/soilDB/misc/fetchNASIS_pedons-flame-graph-time.svg

`simplifyColorData` is now one of the top offenders. It seems that we are running into `ddply`-related lag when there are > 500 or so pedons.

It is not clear if unexported functions like `.formatLandformString` are named in the results.

dylanbeaudette:
Significant progress after conversion from `ddply` &rarr; `aggregate` in `simplifyFragmentData` as of 
02ea69d79b2575e25ef6d77ebc8d800877aff2c7

![image](https://user-images.githubusercontent.com/624277/52003755-42fd0900-247a-11e9-9ac0-bc07c1b187a4.png)

dylanbeaudette:
Significant progress related to `simplifyColorData` via #88. 
![image](https://user-images.githubusercontent.com/624277/52022334-4b226c00-24ad-11e9-8946-9670a54b65a1.png)

dylanbeaudette:
This is likely one of the worst offenders:
`best.tax.data <- ddply(extended_data$taxhistory, 'peiid', .pickBestTaxHistory)`

Conversion to `split` / `lapply` saves 10 seconds on ~ 4k pedons.

dylanbeaudette:
Changes applied and results:
   * ~ 4k pedons in local database SS => 80 seconds (slow simplifyColorData)
   * ~ 4k pedons in local database SS => 44 seconds (fast simplifyColorData)
[...]

PR #187, opened 2021-06-03
`simplifyColorData`, `estimateColorMixture`: use data.table

brownag:
Should make these routines a bit faster. Needs a little more before/after testing, but never pushed this branch to origin.

### DIFF
--- a/R/estimateColorMixture.R
+++ b/R/estimateColorMixture.R
@@ -21,7 +21,7 @@ estimateColorMixture <- function(x, wt='pct', backTransform=FALSE) {
   ## TODO: account for backtransform == TRUE, different return structure
   # sanity check: no NA
   if(any(c(is.na(x$L), is.na(x$A), is.na(x$B)))) {
-    return(data.frame(r=NA, g=NA, b=NA))
+    return(data.frame(r = NA_real_, g = NA_real_, b = NA_real_))
   }
   
   # attempt to fill missing weights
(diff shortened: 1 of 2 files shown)


dylanbeaudette:
Cool. I bet similar code in aqp could use a similar upgrade. Note that I'd like to eventually (soon) replace `soilDB::estimateColorMixture()` with `aqp::mixMunsell(..., method = 'estimate')`. We can do this outside of this PR.

brownag:
In terms of speed a modest (~5 s) improvement with ~4700 pedons mixing about 500 horizons each of moist and dry. Note that this is a "typical" case -- but the data.table optimizations likely offer much more significant benefits when the SPC is large (>10k profiles). Following on from earlier plyr -> base optimization here https://github.com/ncss-tech/soilDB/issues/55

> mixing dry colors ... [422 of 20278 horizons]
> mixing moist colors ... [547 of 23362 horizons]

## Before (current master branch)
image

About 16 seconds (half the execution time) goes to `rgb2munsell` / `farver::compare_colour`.

## After (this PR)

image

Similar breakdown... maybe a little bit faster iteration, but the core code being executed (now by data.table) is unchanged and `rgb2munsell` is still the main workhorse for color code.

---

There are fairly significant differences between the `mixMunsell` and `estimateColorMixture` -- such that the former will not be a drop in replacement within `simplifyColorData` (and all of the table columns that go into fetchNASIS result) without refactoring or writing a wrapper method.

brownag:
Confirmed parity before/after for sRGB hex values in fetchNASIS result, so I will merge this and we can replace the color mixing code with aqp equivalents some time in the future

dylanbeaudette:
Awesome, thanks for all of the benchmarking / testing.

PR #188, opened 2021-06-15
Update `get_soilseries_from_NASIS`

brownag:
- Minor updates to add missing columns to the result of querying the "SC database" table `soilseries` in NASIS.

 - Also handles the possible many:one relationships in mineralogy classes stored in `soilseriestaxmineralogy` by concatenating using the word "over" and ordering based on the `minorder` field

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,5 @@
-# soilDB 2.6.3 (2021-06-11)
+# soilDB 2.6.3 (2021-06-16)
+ * Added new columns to soil classification ("SC") table result of `get_soilseries_from_NASIS()`; now including taxonomic mineralogy class which may contain multiple parts for series with strongly contrasting control sections
  * Updates to `get_SDA_*()` methods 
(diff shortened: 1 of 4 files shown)


dylanbeaudette:
Thanks for the update. 

I completely forgot about this function and need to replace all of my custom code with this as soon as it is merged. I'd like to suggest a `delimeter` argument to `get_soilseries_from_NASIS()` so that mineralogy classes can be collapsed in various ways.

brownag:
> I completely forgot about this function and need to replace all of my custom code with this as soon as it is merged. I'd like to suggest a `delimeter` argument to `get_soilseries_from_NASIS()` so that mineralogy classes can be collapsed in various ways.

I saw the updates you made to the SC db rebuilding code in {SoilTaxonomy}. I was gonna comment and say use soilDB... but then I saw that it didn't handle mineralogy class, and some other columns were not included. Hopefully this will work as a replacement and help with consolidating these queries. Let me know if you have any other ideas you want to see in there

dylanbeaudette:
Yes, thanks for that. The recent additions will make it possible to standardize on the `soilDB` implementation.

Issue #190, opened 2021-06-30
musym sometimes erroneously converted to numeric


kevinwolz:
Here is a simple call to `get_mapunit_from_SDA()` with random mysym values:
`out.multiple <- soilDB::get_mapunit_from_SDA(WHERE = "musym IN ('8E2', '671A')")`

The call results in a data frame with musyms as expected:
```
unique(out.multiple$musym)
[1] "671A" "8E2"
```

However, if we make this same call with just the first musym, returned musym values in the resulting data frame are not correct:
```
out.single <- soilDB::get_mapunit_from_SDA(WHERE = "musym IN ('8E2')")
unique(out.single$musym)
[1] "800"
```

It seems that some point in the process "8E2" is getting erroneously interpreted as scientific notation and converted to numeric. This must not happen when both musyms in the first example are run together because "671A" can not be interpreted as numeric in any standard way, and so the whole musym column remains as character. 

After doing some digging, it looks like this erroneous conversion is done by `utils::type.convert()`, which is called by `.post_process_SDA_result_set()`, which in turn is called as part of the data cleaning process  in `SDA_query()`. 

I don't have a great alternative to propose off hand. It would probably be worth checking to see if any other variables might experience an erroneous conversion like this and select a solution that works across the board. 

How unfortunate that some musyms can be interpreted as scientific notation! I admit that this is likely a very rare case that you supply only these tricky musyms in a query, but, at the same time, musyms that end in "E2" or "E3" are actually quite common.

dylanbeaudette:
That is a tough one. I'm having a hard time finding an alternative to `type.convert()`. Looks like `read.csv()` will do the same thing when there is a column contains a single value that looks like scientific notation.

News to me!
```r
read.csv(textConnection('musym\n"8E2"'))

#   musym
# 1   800
```

Two ideas:
  * keep track of which column contains `musym`, make a copy before `type.convert()` and then replace results if the column classes changes from `character` to `numeric` (feels like a hack)
  * Work with the SDA devs on a new JSON format that *doesn't* quote every column as if it were character (this will take time)

brownag:
[...]

PR #191, opened 2021-07-01
first cut at a solution to #190 and more generally, a formal connection to datatypes as used in SDA

dylanbeaudette:
All tests and R CMD check passing on my local machine. I'm a little concerned that I may have missed an important SQL Server datatype in the conversion code. This style of conversion is not robust to arbitrary use of CAST(xxx AS `SOMEOTHERDATATYPE`) that hasn't been included in the following `swtich()`:
```r
  # parse metadata to create colClasses argument
  cc <- sapply(m, function(j) {
    # extract the data type
    dt <- strsplit(j, split = ',', fixed = TRUE)[[1]][8]
    switch(dt,
           'DataTypeName=char' = 'character',
           'DataTypeName=varchar' = 'character',
           'DataTypeName=nvarchar' = 'character',
           'DataTypeName=datetime' = 'character',
           'DataTypeName=int' = 'integer',
           'DataTypeName=smallint' = 'integer',
           'DataTypeName=real' = 'numeric',
           'DataTypeName=float' = 'numeric',
           'DataTypeName=decimal' = 'numeric'
           )
  })
  
  # convert each column that isn't character
  idx <- which(cc != 'character')
  for(f in idx) {
    df[, f] <- as(df[, f], cc[f])
  }
  ```

Waiting to hear back from Phil on additional datatypes to expect.

@brownag maybe you have a more clever approach.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -8,7 +8,7 @@ importFrom("stats", "aggregate", "complete.cases", "na.omit", "as.formula", "spl
 
 importFrom("utils", "URLencode", "download.file", "object.size",
              "read.csv", "read.table", "setTxtProgressBar",
-             "txtProgressBar", "write.table", "type.convert")
+             "txtProgressBar", "write.table")
 
 importFrom("methods", "slot<-", "as")
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,7 @@
 # soilDB 2.6.3 (2021-06-16)
+ * `SDA_query()` and all functions that call `SDA_query()` get proper column class handling (related to #190), however:
+   - be careful with the use of CAST(): unknown datatypes may not be correctly interpreted
+   - 
  * Added new columns to soil classification ("SC") table result of `get_soilseries_from_NASIS()`; now including taxonomic mineralogy class which may contain multiple parts for series with strongly contrasting control sections
  * Updates to `get_SDA_*()` methods 
    - Extends `get_SDA_property(property = ...)` and `get_SDA_interpretation(rulename = ...)` vectorization over property/rulename to work with any aggregation method. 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by brownag** [comment]
Added _bit_, _numeric_, _text_ and _ntext_ as data type options. I don't know that text and ntext are that prevalent in terms of results but they should be able to be `CAST` to.

I discovered _bit_ by querying the area table and finding that the _obterm_ field was an issue. It is an integer data type that takes values `0`, `1` or `NULL`: https://docs.microsoft.com/en-us/sql/t-sql/data-types/bit-transact-sql?view=sql-server-ver15

```r
soilDB::SDA_query("select top 1 seqnum, areasymbol, areaname, areaacres, obterm, areatypeiidref, areaiid from area")
```

Before adding bit above query gives this error: 
```
[1] "Error in is(object, Class) : \n  trying to get slot \"className\" from an object of a basic class (\"list\") with no slots\n"
attr(,"class")
[1] "try-error"
attr(,"condition")
<simpleError in is(object, Class): trying to get slot "className" from an object of a basic class ("list") with no slots>
```

After: 
``` r
#> single result set, returning a data.frame
#>   seqnum areasymbol   areaname areaacres obterm areatypeiidref areaiid
#> 1     NA      MT001 Beaverhead   3566300      0              4    4802
```

dylanbeaudette:
Thanks for testing + commits. I'll merge after hearing back from Phil Anzel on other possible data types.

dylanbeaudette:
From Kyle, a full list of possible datatypes. I'll add the obvious ones, and we can address the others as we encounter them.

Bigint
Binary
Bit
Char
Datetime
Decimal
Float
Geometry
Image
Int
Money
Nchar
Ntext
Numeric
Nvarchar
Nvarchar(max)
Real
Smalldatetime
Smallint
Smallmoney
SQL_Variant
Text
Timestamp
Tinyint
Uniqueidentifier
Varbinary
Varbinary(max)
Varchar
Varchar(max)
XML
datetime2

PR #199, opened 2021-07-23
fetchNASIS: Implement `fill` argument for NASIS pedons

### LINKED ISSUE #131: test `fill` argument for fetchNASIS() components, consider adding for pedons
Make sure that horizon-less sites don't cause trouble for anything aqp-related.

brownag:
Closes #131; thanks to request from Julie Baker in 2-SON office

Having any depths be `NA` is an aqp/SoilProfileCollection depth logic error. Currently `fetchNASIS()` component filling checks for NA  `chiid` to identify permissible logic errors when `fill=TRUE` and `rmHzErrors=TRUE`. In the case of a "filled" SPC, it is not possible to set the unique horizon ID name because NA values are not allowed. Uses the default `hzID` integer calculated by aqp using ID-topdepth order.

Users of SoilProfileCollections must be aware of the need to check logic of the profiles in the collection before running relevant analyses. They can do this same way as is done in `fetchNASIS()` with `aqp::checkHzDepthLogic()`. Basic SPC operations, including `horizons<-` left join should work correctly in a filled SPC.

https://github.com/ncss-tech/soilDB/commit/d5f39b714f0b2d5d40eb9ad3859bdaf5cb863e70 implements `fill` argument that works for pedon data

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: soilDB
 Type: Package
 Title: Soil Database Interface
-Version: 2.6.3
+Version: 2.6.4
 Authors@R: c(person(given="Dylan", family="Beaudette", role = c("aut"), email = "dylan.beaudette@usda.gov"),
              person(given="Jay", family="Skovlin", role = c("aut")), 
              person(given="Stephen", family="Roecker", role = c("aut")), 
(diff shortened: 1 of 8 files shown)


brownag:
Note that  `rmHzErrors=FALSE` with `fill=TRUE` and horizon data with certain logic errors may create SPCs that do not pass the S4 "valid object" tests in aqp 1.29, resulting in errors that could be difficult to debug. This was resolved in v1.30 which should be released to CRAN soon

Issue #178, opened 2021-03-31
"soil data aggregation engine" vs. wrangling SQL statements or ad hoc R code


dylanbeaudette:
Long-term plan: new macros or UDF to support aggregation on the fly in SQL and parity with WSS output. Contractors are working towards this, but no ETA.

Short-term plans:
  * SQL templates / string interpolation based on SSURGO on Demand / Soil Data Development Toolbox (python)
  * purely R-based solutions (`slab`, `data.table`, etc.) should match SOD / SDDT output when possible

See `misc/soil-data-aggregation` for stubs that can be built-upon for both styles. The (eventual) SDA-based solution will solve most problems, but R-based solutions will be required for data that hasn't been or cannot be pushed to SDA.

brownag:
Here is a small InterpretationEngine proof of concept involving WCS and custom data aggregation I made last week while testing the new R package.

https://github.com/ncss-tech/interpretation-engine/blob/master/misc/demo-SDA-WCS-IE-integration.R

brownag:
#179 adds the ssurgoOnDemand queries in `get_*` method style

I think we could probably get some more stub queries that target some other data categories from here 
https://nasis.sc.egov.usda.gov/NasisReportsWebSite/limsreport.aspx?report_name=SDA-SQL_Library_Home

dylanbeaudette:
Is there ever a valid reason to convert NULL values → 0?

Current SOD:
 * https://github.com/ncss-tech/pysda/blob/8624685f04fef4db0f2d7626b92accc68583a69d/sdaprop.py#L200
 * https://github.com/ncss-tech/pysda/blob/8624685f04fef4db0f2d7626b92accc68583a69d/sdaprop.py#L273

Changes made in #185 remove this step.

Related examples:
```r
library(aqp)
library(soilDB)
library(sf)
library(rasterVis)
library(viridisLite)

# SJQ | Amador Co.

# Mokelumne | Sed. Rock Land
# https://casoilresource.lawr.ucdavis.edu/gmap/?loc=38.46643,-121.02565,z14
bb <- '-121.1099 38.4288,-121.1099 38.4942,-120.9728 38.4942,-120.9728 38.4288,-121.1099 38.4288'
wkt <- sprintf('POLYGON((%s))', bb)

x <- st_as_sfc(wkt)
st_crs(x) <- 4326

(mu <- mukey.wcs(aoi = x, db = 'gnatsgo'))

[...]

Issue #161, opened 2021-01-21
replace plyr / reshape2 functionality with data.table equivalent


dylanbeaudette:
Keeping track of changes / testing related to `data.table` replacement of `plyr` / `reshape` functionality. Each change should be checked and a new test added if there isn't sufficient coverage. Handy bash script:
```
grep 'join(' * | awk '{print "  - [ ] `"$1"`"}' | tr -d ':.R' | sort | uniq
```

### `join`
  - [x] `fetchHenry`
  - [x] `fetchNASIS_pedons`
  - [x] `fetchNASIS_report`
  - [x] `fetchPedonPC`
  - [x] `getHzErrorsNASIS`
  - [x] `getHzErrorsPedonPC`
  - [x] `get_colors_from_pedon_db`
  - [x] `get_component_data_from_NASIS_db`
  - [x] `get_extended_data_from_NASIS_db`
  - [x] `get_hz_data_from_pedon_db`

### `ddply`
  - [x] `fetchHenry`
  - [x] `fetchNASISLabData`
  - [x] `fetchNASIS_components`
  - [x] `fetchNASIS_report`
  - [x] `fetchPedonPC`
  - [x] `fetchSCAN`
  - [x] `getHzErrorsNASIS`
  - [x] `getHzErrorsPedonPC`
  - [x] `get_colors_from_pedon_db`
  - [x] `get_component_from_LIMS`
  - [x] `get_component_from_SDA`
  - [x] `get_phlabresults_data_from_NASIS_db`
  - [x] `simplifyArtifactData`

### `ldply`
  - [x] `SDA-spatial`
  - [x] `fetchSCAN`

### `reshape2::dcast` → `data.table::dcast`
 - [x] `fetchHenry`
 - [x] `simplfyFragmentData`
 - [x] `simplifyArtifactData`

Note that `value` argument becomes `value.var`.

### `melt` → `data.table::melt`
  - [x] `fetchSCAN`

#### Caveats 
   * It appears that `data.table::melt` will not preserve factor type / levels unless asked to do so explicitly via `value.factor` argument. A reasonable solution is to set it after the reshape operation when appropriate.
   * `id.vars` should not contain duplicated column names
   * `melt.data.table` expects that all `measure.vars` are of the same type (e.g. all `numeric`) otherwise warning issued

dylanbeaudette:
Nice work on those last changes. Looks like all that is left is `get_colors_from_pedon_db`?

brownag:
[...]

PR #203, opened 2021-08-06
Add get_SDA_coecoclass

brownag:
### Get mapunit ecological sites from Soil Data Access 
 * Adds `get_SDA_coecoclass()` 

This is an "SOD-style" query that allows for "dominant component" "dominant condition" and "none" aggregation methods applied to component/coecosite tables in SDA. However the aggregation is applied in R.

When `method="Dominant Condition"` an additional field `ecoclasspct_r` is returned in the result with the sum of `comppct_r` that have the dominant condition `ecoclassid`. The component with the greatest `comppct_r` is returned for the `component` and `coecosite` level information. Note that if there are multiple `coecoclasskey` per `ecoclassid` there may be more than one record per component.


brownag:
RE: #161 #178

PR #204, opened 2021-08-06
Add `get_NASISWebReport()` and `get_NASISWebProjects()`

brownag:
* Add 2 new low-level methods (`get_NASISWebReport()` and `get_NASISWebProjects()`) for scraping tables/project IDs from LIMS / NASIS Web Report for support of/comparison with new southwest region project QC/QA tools


dylanbeaudette:
Late in the game, but how do you imagine `get_NASISWebReport()` inheriting, extending, or otherwise interfacing with the less-specific `parseWebReport()`? I have no specific plan for `parseWebReport()`, but would like to see that kind of functionality consolidated into something better than what is currently in place.

brownag:
You are right, there is no need for those methods. `parseWebReport()` is more generic and should cover those use cases.

dylanbeaudette:
Well, no need to trash all of that work--I'm sure that `parseWebReport()` could use some handy wrappers. Not to mention any improvements in the underlying code.

brownag:
I think I would like to revisit the idea again when I refactor the code I had written for those methods to use `parseWebReport()`. 

I had an branch that had not been pushed to the remote, but figured I would push it. Probably should have left it open for discussion rather than merge so quickly as I did a Friday cleanup of my local repo.

PR #221, opened 2021-11-17
Replace rgeos with wk for well-known text processing

### LINKED ISSUE #218: SDA query limits
Recently I fielded some questions regarding SDA queries generated by `SDA_spatialQuery()` for feature classes with several hundred thousand POINT features.

After investigating and checking with Phil Ansel, @dylanbeaudette and I have determined there are limits on the length of input queries to SDA (in addition to the known limits on result)

 - [x] We will be issuing an error for any query that exceeds ~2.5 million bytes (characters) with guidance to chunk up into smaller queries. This provides a little bit of a margin against the known maximum, but exact value pending 

 - [x] Efficiency improvements: In order to enhance efficiency of the queries containing generated WKT strings we will express coordinates at the same precision as the source R object. In order to do this we will convert the SDA WKT processing from using {rgeos} `read`/`writeWKT()` to the processing methods provided by {wk}, as rgeos seems to have fixed (15 decimal place) precision. I tested this out provisionally last March in https://github.com/ncss-tech/soilDB/commit/8eb01fd88fdc46bf80f4eb94f939db6cce18a173 and with minor updates to bring it up to current master branch state this should be good to go.

brownag:
Closes #218

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -13,7 +13,7 @@ License: GPL (>= 3)
 LazyLoad: yes
 Depends: R (>= 3.5.0)
 Imports: aqp, grDevices, graphics, stats, utils, xml2, sp, raster, curl, methods, data.table, DBI
-Suggests: rgdal, jsonlite, httr, sf, rgeos, rvest, odbc, RSQLite, testthat, scales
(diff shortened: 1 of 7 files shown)


dylanbeaudette:
Thanks, looks good. Code seems much simpler with sf + wk.

PR #224, opened 2021-12-09
fetchSDA: components with no horizon data

brownag:
Here is a sample MUKEY (Urban land mapunit) that does not have horizon data. 

``` r
f <- soilDB::fetchSDA("mukey = 412895", duplicates = TRUE)
#> single result set, returning a data.frame
#> single result set, returning a data.frame
#> single result set, returning a data.frame
#> empty result set
#> empty result set
#> Error in UseMethod("within"): no applicable method for 'within' applied to an object of class "NULL"
```

 - Fix is to not apply the chorizon / child table processing code for NULL/ 0-row results. 

 - This will lead to a NULL value to build the SPC from. Therefore we need to change from INNER JOIN to LEFT JOIN for component->chorizon to fill in missing values for potentially missing chorizon records, and handle the cases of missing data before completing the build of SPC. https://github.com/ncss-tech/aqp/pull/245

### DIFF
--- a/R/get_component_from_SDA.R
+++ b/R/get_component_from_SDA.R
@@ -526,7 +526,7 @@ get_chorizon_from_SDA <- function(WHERE = NULL, duplicates = FALSE,
     FROM mapunit) AS mu2 ON mu2.mukey = mu.mukey
    ")},
   "INNER JOIN
-   component    c    ON c.mukey      = mu.mukey   INNER JOIN
+   component    c    ON c.mukey      = mu.mukey   LEFT JOIN
    chorizon     ch   ON ch.cokey     = c.cokey    LEFT OUTER JOIN
    chtexturegrp chtg ON chtg.chkey   = ch.chkey
                      AND rvindicator = 'YES'      LEFT OUTER JOIN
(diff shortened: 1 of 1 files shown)


brownag:
OK, so this will still error `Error: Horizon designation name (hzname) not in horizon names.` without parallel updates to aqp to ensure column names don't get dropped in the NULL case.

```r
soilDB::fetchSDA("mukey = 412895", duplicates = TRUE)
```

With https://github.com/ncss-tech/aqp/pull/245 this returns

``` r
soilDB::fetchSDA("mukey = 412895", duplicates = TRUE)
#> single result set, returning a data.frame
#> single result set, returning a data.frame
#> single result set, returning a data.frame
#> empty result set
#> single result set, returning a data.frame
#> single result set, returning a data.frame
#> empty result set
#> empty result set
#> Warning: all top depths missing from input data
#> converting horizon IDs in column `chkey` to character
#> SoilProfileCollection with 1 profiles and 1 horizons
#> profile ID: cokey  |  horizon ID: chkey 
#> Depth range: Inf - -Inf cm
#> 
#> ----- Horizons (1 / 1 rows  |  10 / 62 columns) -----
#>  hzname    cokey chkey hzdept_r hzdepb_r hzID texture texcl fragvol_l fragvol_r
#>    <NA> 22087546  <NA>       NA       NA    1    <NA>  <NA>        NA        NA
#> 
#> ----- Sites (1 / 1 rows  |  10 / 79 columns) -----
#>     cokey  mukey nationalmusym   compname comppct_r           compkind
#>  22087546 412895          fvn6 Urban land       100 Miscellaneous area
#>  majcompflag localphase drainagecl hydricrating
#>          Yes       <NA>       <NA>     Unranked
#> 
#> Spatial Data:
#>      [,1]
#> [1,]   NA
#> CRS:  NA
```

PR #227, opened 2021-12-21
Missing silent=TRUE in try() call

MatthieuStigler:
Without silent=TRUE, try() will break if there's really an error. So next line (`inherits(d, 'try-error')`) could never be true

### DIFF
--- a/R/SDA_query.R
+++ b/R/SDA_query.R
@@ -200,7 +200,7 @@ SDA_query <- function(q) {
   if (inherits(r.content,'try-error'))
       return(invisible(r.content))
 
-  d <- try(jsonlite::fromJSON(r.content))
+  d <- try(jsonlite::fromJSON(r.content), silent=TRUE)
 
   if (inherits(d, 'try-error'))
     return(invisible(d))


brownag:
> Without silent=TRUE, try() will break if there's really an error. So next line (`inherits(d, 'try-error')`) could never be true

Thanks for reporting this issue. try() is working as expected here. We want the default `silent=FALSE` by the lower-level SDA_query() method so the error message gets printed out to the console and the user is alerted to the fact that they probably are returning a try-error result/something went wrong.

``` r
library(readr)
library(soilDB)

mukey_df <- readr::read_csv("https://raw.githubusercontent.com/MatthieuStigler/Misc/master/dataset_example/mukeys_list.csv")
#> Rows: 25000 Columns: 2
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> dbl (2): mukey, n
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.

vars_cnts <- c("sandtotal_r","silttotal_r","claytotal_r", "awc_r")# "aws050wta")


x <- soilDB::SDA_query(soilDB::get_SDA_property(property = vars_cnts,
                                                method = "Dominant Component (Numeric)", 
                                                mukeys = mukey_df$mukey,
                                                top_depth = 25,
                                                bottom_depth = 50,
                                                query_string = TRUE))
#> Error : lexical error: invalid char in json text.
#>                                        <!DOCTYPE html>  <html lang="en
#>                      (right here) ------^
inherits(x, 'try-error')
#> [1] TRUE
```

SDA_query() will return a try-error if JSON result is not parsed correctly. Higher level methods that use SDA_query() need to make sure that the try error gets handled correctly.

``` r
library(readr)
library(soilDB)

mukey_df <- readr::read_csv("https://raw.githubusercontent.com/MatthieuStigler/Misc/master/dataset_example/mukeys_list.csv")
#> Rows: 25000 Columns: 2
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> dbl (2): mukey, n
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.

vars_cnts <- c("sandtotal_r","silttotal_r","claytotal_r", "awc_r")# "aws050wta")


x <- soilDB::get_SDA_property(property = vars_cnts,
                                                method = "Dominant Component (Numeric)", 
                                                mukeys = mukey_df$mukey,
                                                top_depth = 25,
                                                bottom_depth = 50)
#> Error : lexical error: invalid char in json text.
#>                                        <!DOCTYPE html>  <html lang="en
#>                      (right here) ------^
#> Error: lexical error: invalid char in json text.
#>                                        <!DOCTYPE html>  <html lang="en
#>                      (right here) ------^
inherits(x, 'try-error')
#> Error in inherits(x, "try-error"): object 'x' not found
```
In this case `get_SDA_property()` catches the try error from SDA_query() and `stops()` with the condition... so we get the lexical error printed twice to the console.

https://github.com/ncss-tech/soilDB/blob/e304ef3f4ddadd1435ff537c3afbfcf9713cb8c2/R/SDA_properties.R#L153-L161

brownag:
What might work better here -- to at least make the output from this less repetitive --  is to have SDA_query() take a `silent` argument that is used when it is being called internally (e.g. inside get_SDA_property()). That way the try() output would be suppressed until the end.

MatthieuStigler:
Thanks for the explanation, and sorry for suggesting a wrong patch. I believe though that if `silent=FALSE`, the next line `if (inherits(d, 'try-error'))` will never be true?

brownag:
`silent` only affects whether the error condition is printed out to the console. For example:
``` r
d1 <- try(stop("this is an error"))
#> Error in try(stop("this is an error")) : this is an error
inherits(d1, 'try-error')
#> [1] TRUE

d2 <- try(stop("this is an error"), silent = TRUE)
inherits(d2, 'try-error')
#> [1] TRUE

testthat::expect_equal(attr(d1, 'condition'), 
                       attr(d2, 'condition'))
```

MatthieuStigler:
oh I se, my bad!

PR #244, opened 2022-04-18
downloadSSURGO & createSSURGO: download WSS ZIP archives by soil survey area and create local databases (SQLite/SpatiaLite/Geopackage)

brownag:
A set of tools to download SSURGO ZIP archives from Web Soil Survey containing spatial (.shp) and tabular (.txt) files for individual soil survey areas.

Current plan is to have two methods:
 - `downloadSSURGO()` (previously `get_SSURGO()`) handles downloading/extraction of the source SSURGO exports. This initially will focus on exports available from WSS, as automating Raster Soil Survey downloads from Box is problematic
 - `createSSURGO()` handles building of local databases as SQLite/SpatiaLite/Geopackage from one or more exports. These could have been obtained via `downloadSSURGO()`, directly exported from NASIS or downloaded from other source.
   - This will support vector and raster spatial products (gridded in GPKG only)


dylanbeaudette:
Seems like a fine addition to `soilDB`. 

Note name / functionality overlap with `FedData::get_ssurgo()`. I'd suggest either picking a different name or making it very clear how this is different ("getting") vs. assembling (combining) the results into R objects.

brownag:
> Note name / functionality overlap with `FedData::get_ssurgo()`. I'd suggest either picking a different name or making it very clear how this is different ("getting") vs. assembling (combining) the results into R objects.

There is a capitalization difference so its not really overlapping. I think this name is fine. 

> Also, probably worth adding an argument to toggle download of version that contains the template DB. A lot of potential users of this function will not want the bulkier version with (current) template DB.

Yeah, this is something I was planning on. Also an ability to get just the templates.

The plan here is to additionally provide the methods for building an SQLite instance from one or more ZIPs as well.

dylanbeaudette:
OK. We can figure out the naming later on. I would like to coordinate as best we can since the maintainer of FedData has been really great about working with us.

I like the idea of a function that gets/packages the SSURGO archives into an SQLite DB.

brownag:
Plan is to have two methods:
 - `downloadSSURGO()` (previously `get_SSURGO()`) that handles downloading/extraction of the source SSURGO exports. This initially will focus on exports available from WSS, as automating Raster Soil Survey downloads from Box is problematic
 - `createSSURGO()` that handles building of local databases as SQLite/SpatiaLite/Geopackage from one or more exports. These could have been obtained via `downloadSSURGO()`, directly exported from NASIS or downloaded from other source.
   - This will support vector and raster spatial products (gridded in GPKG only)

brownag:
Sample code for current implementation. Currently supports aggregating multiple survey areas of vector and tabular data, adding primary keys, writing to geopackage etc. 

```r
library(RSQLite)
library(soilDB)

downloadSSURGO("areasymbol IN ('CA067', 'CA077', 'CA632')", destdir = "SSURGO_test")

createSSURGO("test.gpkg", exdir = "SSURGO_test", overwrite = TRUE)

con <- dbConnect(SQLite(), "test.gpkg")  

dbListTables(con)
```

PR #251, opened 2022-06-11
fetchNASIS("components"): use `duplicates` argument and custom profile ID for many:1 relationships of legend, mapunit to datamapunit

brownag:
Many mapunits. Uses `get_component_correlation_data_from_NASIS_db()`, provides `dropAdditional` `dropNonRepresentative` and `duplicates` as optional arguments to `fetchNASIS()`. Creates a non-standard "coiidcmb" from the combination of `lmapunitiid:muiid:dmuiid:coiid` which is used as the result SPC ID column.

This seems to now be working as I intended after checking to make sure all joins have an appropriate ID on left and right hand side.


dylanbeaudette:
Good idea, I've always been annoyed with the previous approach. I'll do some testing when I get back.

brownag:
This works as it should, and does not change _historic_ default behavior. 

Slight breaking change from 2.7.1 in that `duplicates` argument is now being required to load data from mapunit/legend via `fetchNASIS("components')` whereas recently a (possibly incomplete) mapunit and legend table would be joined to SPC result iif the tables were populated in selected set/local DB.

PR #262, opened 2022-08-22
Bump minimum R version due to anonymous function shorthand usage.

cbrueffer:
`r/uncode.R` uses the new `\(x)` anonymous function shorthand introduced in R 4.1, so the minimum R version needs a bump.

Alternatively use of the shorthand could be removed to keep the current minimum version.


brownag:
Dang, this was an oversight. I will correct it by removing the \(x)

Issue #265, opened 2022-09-13
fetchHenry() NA-padding for weekly / monthly granularity


dylanbeaudette:
TODO:
  - [ ] generalize `.fill_missing_days()` with new function and `gran` argument: days, weeks, months
  - [ ] adapt `.formatDates()` with new function
  - [ ] new argument to `fetchHenry()` for generic NA padding
  - [ ] updated docs and tutorial

Further research: https://stackoverflow.com/questions/22439540/how-to-get-week-numbers-from-dates

First approximation here.
```r

.fillMissingGran <- function(x, gran) {
  
  ## TODO this doesn't account for leap-years
  # 366 days
  # 53 weeks
  
  # sequence of possible values
  g.vect <- switch(
    gran,
    'day' = 1:365,
    'week' = 1:52,
    'month' = 1:12
  )
  
  # column to use
  # week / month_numeric are missing
  g.col <- switch(
    gran,
    'day' = 'doy',
    'week' = 'week',
    'month' = 'month_numeric'
  )
  
  # format string
  g.fmt <- switch(
    gran,
    'day' = '%Y %j %H:%M',
    'week' = '%Y %W %H:%M',
    'month' = '%Y %m %H:%M'
  )
  
  
  # add time ID columns as-needed
  # doi is always present
  
  ## "week" not as simple as it seems
  # https://stackoverflow.com/questions/22439540/how-to-get-week-numbers-from-dates
  
  # week
  if(gran == 'week') {
    x$week <- as.integer(format(x$date_time, '%W'))
  }
  
  # month
  if(gran == 'month') {
    x$month_numeric <- as.integer(format(x$date_time, '%m'))
  }
  
  
  # ID missing time IDs
  missing <- which(is.na(match(g.vect, x[[g.col]])))
  
  # short-circuit
  if (length(missing) < 1) {
    return(x)
  }
  
  
  # make fake date-times for missing time IDs
  fake.datetimes <- paste0(x$year[1], ' ', missing, ' 00:00')
  
  # TODO: this will result in timezone specific to locale; 
  #  especially an issue when granularity is less than daily or for large extents
  fake.datetimes <- as.POSIXct(fake.datetimes, format = g.fmt)
  
  # generate DF with missing information
  fake.data <- data.frame(
    sid = x$sid[1],
    date_time = fake.datetimes, 
    year = x$year[1],
    doy = missing.days, 
    month = format(fake.datetimes, "%b")
  )
  
[...]

PR #267, opened 2022-10-04
fetchKSSL: Fix for "status was 'SSL connect error'"

brownag:
Thanks to Ben Marshall @Kramdog for bringing this up / reminding me that the SSL verification was still an issue in some cases. Since jsonlite changed behind the scenes, we need to attend to fetchKSSL-related downloads.

This updates `fetchKSSL()` to use {curl} internally for downloading the gzipped JSON from SoilWeb. This was what was done implicitly w/ `fromJSON()` w/ jsonlite <1.8.1, but with the `ssl_verifyhost` option set in the header.

While the base R approach / default headers included in jsonlite >1.8.1 work fine on non-gov machines we occasionally encounter errors related to SSL verification when in the office or connected to VPN, i.e.
image

Since we recently set up {curl}-based approach for replacing `download.file()` (due to timeout and SSL limitations), we just need to use the standard soilDB curl handle (including  ssl_verifyhost=0) to sort this out. The gzipped json is downloaded to a temporary file with curl, then read as a `gzfile()` connection into `jsonlite::fromJSON()`, and then the temp file is unlinked

### DIFF
--- a/R/fetchKSSL.R
+++ b/R/fetchKSSL.R
@@ -46,13 +46,15 @@
 .getExtended_SoilWeb <- function(f) {
 
   # KSSL geochem, XRD, glass
-  extended.url <- URLencode(paste0('https://casoilresource.lawr.ucdavis.edu/soil_web/kssl/query.php?gzip=1&format=json&what=extended', f))
-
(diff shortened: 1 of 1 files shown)


brownag:
The broken test (https://github.com/ncss-tech/soilDB/actions/runs/3184898261/jobs/5193839591) in R devel is unrelated to the changes. See #265

dylanbeaudette:
Looks good to me, thanks!

Issue #43, opened 2017-12-13
simplfyFragmentData() needs a full review


dylanbeaudette:
This is after switching to `uncode()` in `fetchNASIS_pedons()`.

dylanbeaudette:
The function also needs to be renamed to `simplifyFragmentData`--note the missing "i" in the current function name.

dylanbeaudette:
Note that issues related to fragment content >100% aren't always data errors, they can be related to multiple pedons / KSSL data source (c/o Brian Gardner):

> HI Dylan,
>
> As you recall I was coming up with erroneously high frag contents in some horizons on Kettenbach series.  After a closer look, it seems like the function is being confused because for this series there are two pedons per site. As is practice we have one pedon owned by KSSL that remains unchanging and a second owned by the SSO that is updateable. The updateable one is checked as 'representative pedon'.
>
> This creates a situation where there are two horizons in NASIS with the same horizon labsample number. A quick look at horizons where frags were less than 100% showed several that were still double what was recorded for the horizon in NASIS (for each individual pedon/horizon)
>
> Maybe this is the source of the problem?
>

dylanbeaudette:
This function should also allow for custom specification of fragment classes as an argument.

dylanbeaudette:
Function re-named with backwards compatibility in e2b237d87554fc9c79c87310661909f87fe91ad5.

dylanbeaudette:
Another small issue, `simplifyFragmentData` fails when passed a single row with null `fragvol`:

```r
d <- structure(list(phiid = 1386592L, fragvol = NA_real_, fragsize_l = NA_integer_, 
    fragsize_r = NA_integer_, fragsize_h = NA_integer_, fragshp = structure(NA_integer_, .Label = c("flat", 
    "nonflat"), class = "factor"), fraghard = structure(NA_integer_, .Label = c("noncemented", 
    "indurated", "moderately cemented", "strongly cemented", 
    "weakly cemented", "extremely weakly", "very weakly", "very strongly", 
    "weakly", "moderately", "strongly", "extremely strong", "H", 
    "S"), class = "factor")), .Names = c("phiid", "fragvol", 
"fragsize_l", "fragsize_r", "fragsize_h", "fragshp", "fraghard"
), row.names = 1049L, class = "data.frame")

simplifyFragmentData(d, 'phiid')
```
```
Error: value.var (volume) not found in input
In addition: Warning message:
 Error: value.var (volume) not found in input 
```

[...]

PR #279, opened 2022-11-13
Fix apparent typo in simplifyFragmentData alias

infotroph:
I wasn't sure if this was a typo or an intentional pointer to an old name, but I figured I might as well include the proposed patch in my question -- if intentional, close this freely 😄


infotroph:
🤦 I should have searched harder -- #43 suggests the alias is for historic reasons. Would a note in the docs saying that be helpful?

brownag:
Thanks for pointing this out. 

The function formerly known as `simplfyFragmentData()` was renamed to `simplifyFragmentData()`. There is no object exported with the name `"simplfyFragmentData"`, but `?simplfyFragmentData` will direct to the "corrected" documentation. There is probably no need (even for historic purposes) to retain the alias at this point, but it doesn't really harm anything either.

brownag:
Merging. Thanks also for fixing the broken readme link

PR #297, opened 2023-06-13
.soilDB_curl_get_JSON: suppress messages properly

brownag:
Fixes noisy output for several functions in event of 404 error or similar (non-JSON) result from server. Reported by @dylanbeaudette 

Now `.soilDB_curl_get_JSON()` respects `quiet=TRUE` (as originally intended) and several functions call it "quietly" (seriesExtent, fetchKSSL, fetchOSD, siblings).

### DIFF
--- a/R/AAAA.R
+++ b/R/AAAA.R
@@ -52,18 +52,13 @@ get_soilDB_env <- function() {
 }
 
 #' @importFrom curl curl_download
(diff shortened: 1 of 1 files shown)


dylanbeaudette (approved):

dylanbeaudette:
Thanks, this will be handy for processing large sets of series names that may include inactive / bogus series.

PR #326, opened 2023-12-13
try adding remote soilDBdata reference to DESCRIPTION

brownag:
- mostly just to see whether if in a PR this fixes the visibility of the package within the check environment. not for merge/CRAN.


brownag:
OK, https://github.com/ncss-tech/soilDB/pull/326/commits/c6758bc89254dc397d2340cddcc62ea4f893258c confirmed that Suggests+Remotes soilDBdata works. May be better e CHECK output is same as on CRAN. Can add separate step in current R-CMD-check workflow, or add new workflow, for extended data checks--which need to be made better anyway.

PR #351, opened 2024-05-17
SDA_query: handle API failures with more grace

brownag:
Soil Data Access goes down intermittently for maintenance or due to crashes of the server. This PR adds handling to check the response content type for `SDA_query()` POST requests.

Essentially, any content type other than JSON (data) or XML (error message) indicates the service is temporarily down. When the service is down it does not produce normal response that can be used by either the data or error parsing approach. 

As an additional item in this PR I will develop a wrapper function that can catch errors and print an informative message for the user--there are several instances of functions that use `SDA_query()` but do not handle the possibility of try-error response--which can result in some unusual error messages in lieu of expected result. We will distinguish this "service is down" case from other possible issues (say, incorrectly formatted arguments that affect the query used)


dylanbeaudette:
Thanks. The additional error handling will help downstream functions fail gracefully. I never expected that SDA would be "down" so frequently.

PR #352, opened 2024-05-17
createSSURGO: add support for creating DuckDB and other DBI-compatible databases

brownag:
For a few months I have been exploring the idea of making SSURGO DuckDB databases. 

To further support incorporating this into my workflows, this PR configures `createSSURGO()` to optionally use [{duckdb}](https://duckdb.org/docs/api/r.html) instead of {sf}+{RSQLite}. To toggle on this behavior use `createSSURGO(..., duckdb=TRUE)`.

In the future I would like to develop benchmarks and examples for large database processing--i.e. all of SSURGO in DuckDB vs. Spatialite or Geopackage.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -13,7 +13,7 @@ License: GPL (>= 3)
 LazyLoad: yes
 Depends: R (>= 3.5.0)
 Imports: grDevices, graphics, stats, utils, methods, aqp (>= 2.0.2), data.table, DBI, curl
-Suggests: jsonlite, xml2, httr, rvest, odbc, RSQLite, sf, wk, terra, raster, knitr, rmarkdown, testthat
+Suggests: jsonlite, xml2, httr, rvest, odbc, RSQLite, duckdb, sf, wk, terra, raster, knitr, rmarkdown, testthat
 Repository: CRAN
 URL: https://ncss-tech.github.io/soilDB/, https://ncss-tech.github.io/AQP/
 BugReports: https://github.com/ncss-tech/soilDB/issues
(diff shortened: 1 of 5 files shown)


dylanbeaudette:
Good idea. I haven't tried, but there is certainly a lot of noise on the internet about how great duckdb might be. I'd also suggest unzipping SSURGO archives in parallel--that seems to be the biggest bottleneck right now.

brownag:
> Good idea. I haven't tried, but there is certainly a lot of noise on the internet about how great duckdb might be.

It is definitely very powerful.

Right now a slight barrier to my use of it for spatial stuff is interchanging between the internal duckdb format and WKB for use in sf or similar. Can use a query that invokes ST_AsWKB(), or there is a wrapper package {duckdbfs} to do this

> I'd also suggest unzipping SSURGO archives in parallel--that seems to be the biggest bottleneck right now.

That is an improvement we could make for `downloadSSURGO()`, for sure. I am also curious how parallel unzip would compare to direct read with GDAL /vsizip/

brownag:
Instead of `duckdb=TRUE` argument, consider initializing a DBIConnection object, the class of which can be used to trigger driver specific routines. This would likely make the routine more easily generalizable beyond RSQLite and duckdb

brownag:
In a0915164 I have implemented a more generalized approach using an optional connection (`conn`) argument. 

Current pattern of specifying `filename` can still be used for GPKG or SQLite output, or an existing connection can be provided. 

DuckDB now requires a custom connection be provided, and gets some special handling due to the unique way spatial data are handled internally. 

For all other DBI connection types the spatial data are written with sf. I have tested this with PostGIS only at this time (see 87ff3917 for minimal example), but in theory the routine should work with any DBI-compatible driver that supports spatial data.

Issue #61, opened 2018-03-08
SCAN/SNOTEL metadata


dylanbeaudette:
Should be possible to snag it here:

![image](https://user-images.githubusercontent.com/624277/37181865-907b3aba-22e3-11e8-8257-0b7d016a41ad.png)

Asking Deb Harms to be sure.

dylanbeaudette:
Seems to work. Still need to test, document, and then explain in the related tutorial:

http://ncss-tech.github.io/AQP/soilDB/fetchSCAN-demo.html

@brownag : figured this would be a nice addition to CDEC-based eval of climate data in our part of the world.

Example:
```r
# bleeding-edge version
library(soilDB)
x <- fetchSCAN(site.code=c(462, 697, 574), year=2016)

# new metadata element in list
x$metadata
            Name Site      State Network County Elevation_ft Latitude Longitude          HUC
275 Ebbetts Pass  462 California  SNOTEL Alpine         8661 38.54970 -119.8047 160502010104
507 Leavitt Lake  574 California  SNOTEL   Mono         9604 38.27594 -119.6128 160503020104
741  Poison Flat  697 California  SNOTEL Alpine         7736 38.50576 -119.6262 160502010102

# all stations can be seen with
data(SCAN_SNOTEL_metadata)

# or
SCAN_site_metadata(site.code)
```

dylanbeaudette:
Another reminder to link back to the great work in https://github.com/mrke/NicheMapR

dylanbeaudette:
Still waiting to hear back from Deb on the mapping between SCAN/SNOTEL site code and KSSL pedon ID.

jskovlin:
I just looked into this and the SCAN/SNOTEL station ID should map back to pedon ID will be via the site.climstaid field in the associated lab pedons.  Will need to add that field to fetchNASIS/fetchKSSL. Probably easier to track them all back and save the corresponding pedon ID's in the scan metadata file in the package for easy access.  One thing I'm not sure about is if the site.climstaid field has been consistently populated, my initial look indicates that it might not be.  However, there are only just over a 1000 stations and not all of them may have corresponding lab data.

dylanbeaudette:
@jskovlin Excellent detective work! I knew that there was a linkage somewhere in one of our databases. Do you have an example sites in mind that I can use to test? What we need is a NASIS or LIMS style query that pulls down the latest site / pedon records that have a non-null `site.climstaid` field.

jskovlin:
[...]

Issue #184, opened 2021-05-18
`fetchSCAN` upgrades


brownag:
- [x] handle error on timeout etc. gracefully and check other error handling logic is applied consistently (empty objects, try-error result)
- [x] refactor code involving `melt` `ddply` etc #161
- [ ] consider optimizations for requests to [AWDB (Air-Water Database) Web Service](https://www.nrcs.usda.gov/wps/portal/wcc/home/dataaccess/helpCenters/!ut/p/z1/rZJNb8IwDIb_Sjn0WNm0KUmPXQ-DFQkxYKW5TCEY2q1fQES3f7-CdtkBuknzzcpr57Ufg4Q1yEqd870yeV2postTOXoVImLDGcMpiomL82ges5mPHsYMkqtg4opwLBYYP0aMYRhEfDldcIyXI5C_qccbEWJf_QtIkLoyjckgbbW2dF0ZqoyNWV2SjVtllKW0ptPJyqhobGxpY53oeM41_Uiuz5dujc63kHKXcaE9cgQn32E7NXRUQBsHA-4HjCka0ebb_R178v5wyeW_nvmvglsLfugTdATSziS_bZJDcs6phVVVH8sO-eKPOxgjPPVR6s4ofzscZNixuvD5MLD-B1hNuVqVwvt03p9Fu9xlxX4w-AKbGhVs/dz/d5/L2dBISEvZ0FBIS9nQSEh/)
- [ ] address multiple sensors #14
- [x] Add support for alternate time intervals (e.g. `'hourly'`) and other custom parameters in batch interface (#260)
- [x] Add StationMetadata.stationDataTimeZone to SCAN_SNOTEL_metadata object
- [x] Add support for SNOWLITE stations

brownag:
A note that now with explicit support for hourly data we should be able to put those data into correct timezone to make accurate datetime objects for data from across multiple timezones. 

The StationMetadata should include `stationDataTimeZone ` attribute which can be added to cached `SCAN_SNOTEL_metadata` data object in the package rather than queried on the fly. This value can be used inside the fetchSCAN internal methods for conversion of each sensor time series within a site

<html>
<body>


stationDataTimeZone | BigDecimal | The time zone in which the station collects data.
-- | -- | --



</body>
</html>

dylanbeaudette:
An example:

```
[...]

PR #357, opened 2024-09-18
Scan metadata updates

dylanbeaudette:
This PR addresses several TODO items for `fetchSCAN()` and related functions:
  * #61 -- new site metadata, derived from authoritative sources that have been updated / corrected over the last 2 years
  * #184 -- timezone is now available in the station metadata, and a new `datetime` column initialized as timestamp with timezone

The latest metadata creation code and intermediate files are at `data-raw/scan-snotel-current`. This code and resulting data replace all previous efforts to combine WCIS (station information) and LDM (user pedon ID and pedlabsampnum) sources.


brownag:
Looks good to merge, with one suggestion for a comment above.

I think we can now close #61 as the new data-raw script should allow us to keep up with any new  updates made in linkages between WCIS and LDM. I have linked that issue to this PR per the comment in script, unlink it if you don't think issue is fully resolved.

Not sure whether there will be efforts to characterize sites that lack pedons, or if there is more detective work needed to link more pedons, but if so, from {soilDB} perspective, looks like @jskovlin and @dylanbeaudette have "solved" it and we can now just re-run update script periodically.

I was curious how many sites still lack linked pedon data, see below.

``` r
library(soilDB)
x <- SCAN_site_metadata()
aggregate(
  x$upedonid,
  by = list(Network = x$Network),
  FUN = \(i) c(no_pedon = sum(is.na(i)), n = length(i))
)
#>   Network x.no_pedon x.n
#> 1   CSCAN         20  21
#> 2    SCAN         45 210
#> 3    SNTL        689 910
#> 4   SNTLT         42  45
```

Also probably pretty close to closing #184 with timezone and SNOWLITE updates. Not sure if there is much we want to do with the remaining TODOs in that issue.

dylanbeaudette:
> Looks good to merge, with one suggestion for a comment above.
> 
> I think we can now close #61 as the new data-raw script should allow us to keep up with any new updates made in linkages between WCIS and LDM. I have linked that issue to this PR per the comment in script, unlink it if you don't think issue is fully resolved.

Thanks for looking over the PR and commentary. I agree that we can close #61, with the understanding that there is still work to be done fixing station information in WCIS and in NASIS. Jay is working on a batch of those right now.
 
> Not sure whether there will be efforts to characterize sites that lack pedons, or if there is more detective work needed to link more pedons, but if so, from {soilDB} perspective, looks like @jskovlin and @dylanbeaudette have "solved" it and we can now just re-run update script periodically.
> 

Yes, it will get progressively "better".


> Also probably pretty close to closing #184 with timezone and SNOWLITE updates. Not sure if there is much we want to do with the remaining TODOs in that issue.

I'll close it, and open a new ticket for the question about multiple timezones.

dylanbeaudette:
> Also, no issue with this that I can see, but just a thought: when combining hourly data for stations in multiple time zones the `datetime` column in resulting sensor data uses the time zone from the _first_ station.
> 
> Dissimilar timezones appear to be correctly converted when the data are combined--but curious if we want to give the user the ability to specify a target time zone explicitly (rather than implicitly through the choice of their first station ID).

Good points, I hadn't thought about that carefully and there are no tests in place to ensure that data are reasonable. I also hadn't considered what "happens" when combining multiple timezones. I had to convince myself that it was reasonable, and it is, just as you describe:

```r
x <- c(
  as.POSIXct('2024-01-01 12:00:00', tz = 'etc/gmt+8'),
  as.POSIXct('2024-01-01 12:00:00', tz = 'etc/gmt+10')
)

# correctly converted to single TZ
x

# new TZ is the first one
table(format(x, format = '%Z'))
```

I'll open a new issue to implement selection of a timezone. I'll also add a note in the manual + tutorial.

dylanbeaudette:
Multiple TZ is slightly more complicated than I thought, as until now, the ordering of stations in the results was the numerically-sorted `site.code`. Just fixed this.

brownag:
> Multiple TZ is slightly more complicated than I thought, as until now, the ordering of stations in the results was the numerically-sorted `site.code`. Just fixed this.

See above review comment on what you just deleted, I think that you need to check for `NULL` site code before preserving user input.

brownag:
After discussion on above test added in https://github.com/ncss-tech/soilDB/pull/357/commits/782288996686d7942f2f8ec12a0e1697b189f093#diff-129da79a9437e382a28369436cd7042f113ce828f5a421f7132d59f2cfbded0d which failed on Linux I implemented standardization of the timezone based on `tz` argument to `fetchSCAN()` in https://github.com/ncss-tech/soilDB/pull/357/commits/f762612952f0ba128c35dbe2b752bf8a2350944f. This change closes #358. 

The default behavior ( `tz="US/Central"`) converts all times to central time, to which is a logical choice for a network that is US based. However, this default behavior will result in a mixture of times relative to UTC: CDT (-0500) and CST (-0600) depending on whether dates in the time series occur during daylight savings time. While the offsets differ, times are correct and the results are  tested and consistent across platforms. 

To ensure that a common offset is returned, users need to specify `tz="UTC"` or use a timezone that does not use daylight savings (e.g. `tz="US/Arizona"` which is always -0700 relative to UTC)

brownag:
Just pushed a small change in https://github.com/ncss-tech/soilDB/pull/357/commits/d96cf6f9858255cacdee3c48612801253bb68d58 due to an oversight I made, the issue was only evident for `timestep="Hourly"`

dylanbeaudette:
Thanks for the more nuanced solution to the TZ and multiple above ground sensor problems. Are we good to merge ?

brownag:
> Thanks for the more nuanced solution to the TZ and multiple above ground sensor problems. Are we good to merge ?

Yep, I think this is good to go in that it does what is expected. The only remaining "issue" is perhaps the choice of default timezone.

I noticed that with the default "US/Central" timezone (or any timezone that observes DST) you will get 2 records for the same hour slot when `timeseries="Hourly"` and the time "falls back" in November. This I think is "expected" behavior if we are allowing for conversion to any local timezone (be it user specified or specific to a group of stations). 

To get stability in times and DST need to use "UTC" or a specific offset as you had originally set up.

The caveat to specific offset being that the exact output will differ based on platform. It seems that on the Linux systems tested (recent Ubuntu and Manjaro) that the offset gets applied to the time (converted to UTC), whereas on Windows and Mac the offset is preserved with the local time. Either is a "correct" representation of time, but final output varies with platform. 

Choosing a consistent timezone default at least prevents multiple queries across different extents or different groups of stations from returning different results silently. I suppose this is not a huge problem because if combining POSIXct results from multiple queries the first timezone is used. We probably want to make sure the tutorials cover this in some regard. The `fetchSCAN()` docs currently do mention it effect of DST and describe how to use `tz` argument for `datetime` column result. I just updated the example output to show the new behavior and an example of correcting using a specific stations metadata.

brownag:
Do we want to add a message that indicates to users requests that span multiple `dataTimeZone` and/or if the dataTimeZone is outside the defaults of US/Central (i.e. not -5 or -6)? This might help alert to the need to specify `tz` argument to a more locally relevant value

dylanbeaudette:
> Do we want to add a message that indicates to users requests that span multiple `dataTimeZone` and/or if the `dataTimeZone` is outside the defaults of US/Central (i.e. not -5 or -6)? This might help alert to the need to specify `tz` argument to a more locally relevant value

Yes, a message when >1 TZ result from the selected stations is a good idea. I can do that this morning if you are busy.

brownag:
> > Do we want to add a message that indicates to users requests that span multiple `dataTimeZone` and/or if the `dataTimeZone` is outside the defaults of US/Central (i.e. not -5 or -6)? This might help alert to the need to specify `tz` argument to a more locally relevant value
> 
> Yes, a message when >1 TZ result from the selected stations is a good idea. I can do that this morning if you are busy.

Did it yesterday, just didnt push. See https://github.com/ncss-tech/soilDB/pull/357/commits/309e3571c91d8b4098053c86f643caa2c447ec45

PR #366, opened 2024-11-18
fetchVegdata: add siteobsiid to vegplot tables; activate vegbasalarea table

natearoe:
siteobsiid absent from a couple tables, including vegplot. Its inclusion should make joining to siteobs easier. 

get_vegplot_speciesbasalarea_from_NASIS() was built in get_vegplot_data_from_NASIS_db.R, but needed to be referenced in fetchVegdata - added this reference.


brownag:
Looks good to me, thanks for catching this and making the results more consistent

PR #373, opened 2025-02-22
Vegplot Transect: Add support for Transect Ground Surface Cover

brownag:
Add `get_vegplot_groundsurface_from_NASIS_db()` <strike>and `get_vegplot_transplantfoliar_from_NASIS_db()`</strike> from @phytoclast

Following from discussion in the FPAC-NRCS R User Group Teams channel.

- [x] I have tentatively added `get_vegplot_BA_snags_from_NASIS_db()` but I think it could potentially be combined with existing functions 
- [x] Documentation could be cleaned up and made more consistent across all vegplot vegetation data functions
- [x] <strike>`si` argument for converting units could be implemented for other pre-existing functions and columns</strike> See https://github.com/ncss-tech/soilDB/issues/374


brownag:
Confirmed the existing query in `get_vegplot_from_NASIS_db()` already includes siteiid, siteobsiid, vegplotid, vegplotname, overstorycancontotalpct, basalareaplottotal, treesnagdensityhard, treesnagdensitysoft which are all 1:1 with vegplot table, so `get_vegplot_BA_snags_from_NASIS_db()` can safely be omitted from this PR

brownag:
Also `get_vegplot_transpecies_from_NASIS_db()` already covers the data elements returned by `get_vegplot_transplantfoliar_from_NASIS_db()`. 

The unit conversion portion would be valuable to incorporate--need to look at all the columns that need SI unit conversions to handle it more generally.

phytoclast:
Fair enough. I agree that the fetch function probably should not auto convert units and obscure stored values. I agree it would be wiser to do this in a separate deliberate action. The rounding that I employ after conversions is to maintain the level of precision in the original field data. Plant height can be recorded to the nearest tenth of a foot, which means measurement in meters can only be considered to the nearest decimeter. Diameter can be recorded to the nearest tenth inch, meaning I can only read d-tape to the nearest cm.

I do already have tools in vegnasis package that consolidates the major variables and applies conversations so that data from multiple sources outside NASIS can conceivably be combined for analysis. Veg data collected by NPS, Natureserve, and state natural heritage programs are recorded in SI units, consistent with vegetation literature. Likewise, the calculations used in the background for scaling vegetation graphs are done more easily in SI because land area, stem diameter, and height are all in the same base units. Conversion or reversion to USC units for end user reporting should, therefore, be done after data is aggregated and summarized.

brownag:
`get_vegplot_groundsurface_from_NASIS_db()` has been added to the vegplot data returned by `fetchVegdata()` and tested for basic functionality via the [soilDBdata snapshot](https://github.com/brownag/soilDBdata)

PR #376, opened 2025-03-01
`fetchSOLUS`: Add mpspline based interpolation methods

brownag:
I was going through and deleting old branches and saw this one which I never merged.

This PR would add `mpspline2` to suggests, and add several variants of equal-area spline depth interpolation `method` options for  SoilProfileCollection conversion in `fetchSOLUS()`

At the time I considered that the approaches I developed for depth interpolation data should probably be generalized to a function in aqp that was data source agnostic. aqp already brings in mpspline2, and the general functionality would be more useful there, which is probably why I did not pursue this PR further. Opening a draft PR to remind myself to look back at this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -13,7 +13,7 @@ License: GPL (>= 3)
 LazyLoad: yes
 Depends: R (>= 3.5.0)
 Imports: grDevices, graphics, stats, utils, methods, aqp (>= 2.0.2), data.table, DBI, curl
-Suggests: jsonlite, xml2, httr, rvest, odbc, RSQLite, sf, wk, terra, raster, knitr, rmarkdown, testthat
+Suggests: jsonlite, xml2, httr, rvest, odbc, RSQLite, sf, wk, terra, raster, mpspline2, knitr, rmarkdown, testthat
 Repository: CRAN
 URL: https://github.com/ncss-tech/soilDB/, https://ncss-tech.github.io/soilDB/, https://ncss-tech.github.io/AQP/
 BugReports: https://github.com/ncss-tech/soilDB/issues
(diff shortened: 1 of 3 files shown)


brownag:
This feature can likely be simplified (or is perhaps redundant) now that mpspline2 (>=0.1.9) supports splining multiple variables in one call. 

The SPC output from `fetchSOLUS()` can directly be directly passed to aqp `spc2mpspline()` (requires aqp >=2.3) to achieve the same result, with more options for user control :

``` r
library(aqp)
#> This is aqp 2.3
library(soilDB)

b <- c(-119.747629, -119.67935, 36.912019, 36.944987)

bbox.sp <- sf::st_as_sf(wk::rct(
  xmin = b[1], xmax = b[2], ymin = b[3], ymax = b[4],
  crs = sf::st_crs(4326)
))

ssurgo.geom <- soilDB::SDA_spatialQuery(
  bbox.sp,
  what = 'mupolygon',
  db = 'SSURGO',
  geomIntersection = TRUE
)

vn <- c("sandtotal", "silttotal", "claytotal", "cec7")
res <- fetchSOLUS(
  ssurgo.geom,
  depth_slices = c("0", "5", "15", "30", "60", "100", "150"),
  variables = c("sandtotal", "silttotal", "claytotal", "cec7"),
  output_type = "prediction",
  method = "step",
  grid = FALSE,
  samples = 10
)
#> Loading required namespace: rvest
#> Loading required namespace: terra
#> NOTE: SOLUS predictions represent depth slices (method="slice")
#> Consider using method="constant" or method="linear".

out <- spc2mpspline(res, var_name = paste0(vn, "_p"))
#> Loading required namespace: mpspline2
#> 1 depth range(s) with missing analytical data removed from 1.
# ... omitted for brevity ...
#> 1 depth range(s) with missing analytical data removed from 8.
plot(out, divide.hz = FALSE, color = "claytotal_p_spline")
```



Probably the data.table implementation from this PR is quite a bit faster when dealing with many SOLUS "profiles" than the mpspline2 base R iteration... but the above works fine.

Issue #250, opened 2022-05-13
Implement SQLite/T-SQL compatible syntax for `get_SDA_*()` methods


brownag:
- [x] 1. Why is SQLite GROUP_CONCAT() so slow when run locally?
   - resolved by adding indices to e.g. `createSSURGO()`
  
 2. Generalize temporary table conversion
   - [x] `get_SDA_property()`: dominant component (numeric), weighted average, min/max aggregation
   - [x] `get_SDA_interpretation()`: dominant component (numeric) and weighted average, min/max aggregation
   - [x] `get_SDA_hydric()`: all methods

Further development following https://github.com/ncss-tech/soilDB/pull/246;  I think I have figured out an approach to handle TODO item 2, that with indexing of the temp tables may resolve item 1. 

EDIT: indexing of the tables alone resolves item 1, temporary tables are a much more involved fix for many different queries

brownag:
At long last, in #381, I've developed a scheme for dealing with item 2 (part 1) using CTEs. 

This applies to certain aggregation methods in `get_SDA_property()`, `get_SDA_interpretation()` as well as `get_component_from_SDA()`.

PR #381, opened 2025-03-10
get_SDA_property: CTEs for SQLite compatibility and query readability

brownag:
This PR replaces `SELECT ... INTO ...` SQL syntax with Common Table Expressions (CTEs) in `get_SDA_property()`. Temporary tables are used in the dominant component (numeric) and weighted average aggregation methods.

 - [x] implement CTEs for Weighted Average aggregation (includes dominant component, numeric)
 - [x] implement CTEs for MIN/MAX aggregation
 - [x] Benchmark against gNATSGO and SSURGOPortal output for a few other properties (e.g. available water capacity)

This change has the benefit of increasing the compatibility of `get_SDA_property()` methods with SQLite and other dialects. I've done some benchmarking of old vs. new as well as against SSURGOPortal output. We have 1:1 parity across tools and dialects, and all existing soilDB tests are passing with the new code.

In this PR I've also built out the infrastructure that will make it relatively easy implement this approach for the other areas I identified in #250.

Before merging I will be testing this across the country to verify there aren't edge cases when querying multiple properties with very different patterns of missing values, and adding additional tests if needed to make sure that type of edge-case functionality is fully tested.

### DIFF
--- a/R/get_SDA_property.R
+++ b/R/get_SDA_property.R
@@ -358,7 +358,7 @@ get_SDA_property <-
           property, property, property)
   }
   
-  .property_min_max_CTE <- function(property, top_depth, bottom_depth, WHERE, FUN, include_minors = TRUE, miscellaneous_areas = FALSE) {
+  .property_min_max_CTE <- function(property, top_depth, bottom_depth, WHERE, FUN, include_minors = TRUE, miscellaneous_areas = FALSE, sqlite_dialect = FALSE) {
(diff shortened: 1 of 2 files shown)


brownag:
This looks good to go. We have good parity with prior existing queries in soilDB, as well as SSURGOPortal. The queries run and are consistent using both SDA and local SQLite sources (i.e. created with `createSSURGO()` or with SSURGOPortal)

PR #393, opened 2025-04-25
Add `get_phroots_from_NASIS_db()` for querying and formatting data from phroots table

dylanbeaudette:
Just a simple function with a single task. Data are returned with pedon and pedon horizon record IDs for linking to output from `fetchNASIS()` and related.


R/get_phroots_from_NASIS_db.R
```diff
@@ -0,0 +1,54 @@
...
+    return(NULL)
+  }
+  
+  # toggle selected set vs. local DB
+  if (SS == FALSE) {
+    q <- gsub(pattern = '_View_1', replacement = '', x = q, fixed = TRUE)
+  }
+  
+  # exec queries
+  d <- dbQueryNASIS(channel, q, close = FALSE)
```
brownag, line 40: We do not need to leave the connection open here.

 `dbQueryNASIS()` and `uncode()` default behavior is to close connections when attribute "isUserDefined" is FALSE (i.e. when `dsn` is `NULL`). 

```suggestion
  d <- dbQueryNASIS(channel, q)
 ```

File: R/get_phroots_from_NASIS_db.R [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+  
+  # convert coded -> text/factor representation of values
+  d <- uncode(d, dsn = dsn)
+  
+  # encode ordered factors
+  d$rootsquantityclass <- factor(
```
brownag, line 46: `rootsquantityclass` is not currently a "ranked" domain, so I don't think we should be manually encoding it as an ordered factor. But the choice list does have a logical sequence, and that matches what you have done here. Possibly we should update this in NASIS data model--would be a simple fix that shouldn't affect anything.

brownag: `NASISChoiceList()` should be used here

```suggestion
 d$rootsquantityclass <- NASISChoiceList(d["rootsquantityclass"])
```

R/get_phroots_from_NASIS_db.R
```diff
@@ -0,0 +1,54 @@
...
+    d$rootsquantityclass, levels = c('very few', 'moderately few', 'few', 'common', 'many'),
```
brownag, line 47: Use domain information directly from NASIS rather than hard-coding
```suggestion
```

R/get_phroots_from_NASIS_db.R
```diff
@@ -0,0 +1,54 @@
...
+    ordered = TRUE
```
brownag, line 48: Default behavior of `NASISChoiceList()` is `ordered=TRUE` to use the domain ranking (if it exists)

```suggestion
```

R/get_phroots_from_NASIS_db.R
```diff
@@ -0,0 +1,54 @@
...
+  )
```
brownag, line 49:
```suggestion
```

brownag (changes requested):
Cool. Thanks.

I've made a few minor suggestions in my review.

In addition I think we should provide an aggregation method to produce a data.frame result that is suitable for directly joining to horizons by `phiid`, but we can add that after this is merged if  you don't already have a method worked up.

dylanbeaudette (commented):
Thanks for the suggestions, I've accepted all suggested changes. Since there are many possible ways to aggregate these data, I'd suggest saving that for "later".

brownag (approved):

PR #395, opened 2025-06-30
Linting 1

brownag:
Some linting with {flir} and general cleanup. First steps in some big organizational work I am planning for soilDB 2.9.x

CI tests are passing locally, but need to run a few additional tests on alternative data sources, NASIS etc.


dylanbeaudette:
Thanks, there are a lot of tedious fixes (spaces around `=`) addressed here. I've gone through about half of the files and most look reasonable. I do worry a little bit about changes to coding style--where those performed automatically by {flir}? For example unnesting of `if()` statements, moving assignment from ` -> .` to `. <-`, and substitution of `anyNA()` for `if(any())`. These all seem rational and improve code readability to me. I'd like to spend some more time ensuring that there aren't subtle differences in behavior with NA or unusual values. 

In general, changes to coding style makes me want more tests.

brownag:
These were almost all done manually by me after they were flagged by flir, or identified while I was in there fixing other things. 

I don't really need any assistance checking, but if you have some tests that you want to add feel free to add them.

brownag:
I've now done some additional testing of functions that lack unit tests and feel pretty confident this is good to go. If any issues come up we will fix them and add appropriate tests so they don't break again!

PR #403, opened 2025-08-13
SSURGO Dominant Ecological Site vignette

brownag:
This is a draft vignette covering use of Soil Data Access to obtain spatial and tabular data, and the specific use case of creating thematic maps of ecological sites with `SDA_spatialQuery()`, `fetchSDA_spatial()`, and `get_SDA_coecoclass()`
 - [x] Link out to more related soilDB functions (e.g. `SDA_query()`)
 - [x] Expand on use of local data sources (link created to `createSSURGO()` vignette)
 
 for #157

Issue #157, opened 2021-01-16
soilDB vignettes / bookdown chapters


brownag:
General TODOs suggested in #124 
 - inventory soilDB related tutorials, examples, etc.
 - organize soilDB content via bookdown or related system 
 - provide links to extended examples in manual pages
 
I think a good goal for soilDB in 2021 is to target creation of three or four different [vignettes](https://r-pkgs.org/vignettes.html) (long-form documentation) or bookdown "chapters".

These can be linked to relevant helpfiles for methods, allowing for more concise help but also more complete examples.

This will be a good exercise for all contributors and future contributors, as we focus our interfaces and work to provide even more comprehensive and generic coverage in a future soilDB 3.0.

**Some ideas** might be better as "vignettes" -- more document/tutorial like -- others could be more open-ended challenge exercises (with something resembling a "right answer" at least for the supplied example data)

 - SDA Components -- Accessing SSURGO and STATSGO data from Soil Data Access
 
   - Prequel to "tabular aggregation" with some future aggregation-engine 
  
 - SDA Spatial -- Points, polygons and bounding boxes from SSURGO, STATSGO, and survey area polygons fetched from Soil Data Access, 
   - `SDA_spatialQuery` and `fetchSDA_spatial`.
 
 - Soil Survey Timelines -- A walk-through of @smroecker's soil survey area timeline, pedons over time, etc. graphs and datasets. 
 
 - Web Coverage Service -- SoilWeb web coverage service (raster) demonstration ISSR800 and SSURGO MUKEY
 
 - SoilWeb "Series Concept" tools -- e.g. `fetchOSD`, `seriesExtent`
 
 - Thematic maps of soil taxonomic concepts with {soilDB} and {tmap} - @jskovlin 
 
 - NASIS Pedons & NASIS Components -- two separate guides showing a "real" fetchNASIS dataset, with problems to solve; all reproducible because it will be a [reduced] SQLite source (2.6.x+); exercise objective is to create a variety of reports or output that show the errors (say, to hand off to some hypothetical person to fix)
 
 - Custom NASIS queries and NASIS automation with {soilDB}

brownag:
These are some links to tutorials and demos gleaned from the soilDB docs whilst doing updates to \href syntax for .Rd to Roxygen conversion. Seems like as good of a place as any to keep these links for now. 

[...]

PR #404, opened 2025-08-13
Local SSURGO databases vignette

brownag:
- Adds draft vignette expanding on local rather than SDA sources of SSURGO data
 - For TODO item number 2 from #403 and for #157


brownag:
This should be a decent start on the topic. Will probably make some revisions in not too distant future.

Issue #375, opened 2025-02-24
`ISSR800.wcs()` `mukey.wcs()` and `seriesExtent()` terra usage is not robust to missing/invalid GDAL/PROJ installation


brownag:
Make tests more robust to potentially missing proj.db. May be related to use of Windows binaries from the RSPM as opposed to CRAN?

e.g. tests occasionally generate errors and warnings on Windows: https://github.com/ncss-tech/soilDB/actions/runs/13509235493/job/37745790683

```
PROJ: proj_create_from_database: Cannot find proj.db (GDAL error 1)
```

and 

```
Error: Error: [rast] empty srs
```

brownag:
So, those skip commands are not adequate to catch this condition, they were already present and the error still (intermittently) happened. sf/terra _are_ installed but the PROJ database cannot be found due to not being configured properly This type of thing happens when using the RStudio package manager, possibly also with sources like R-Universe or systems with multiple PROJ installs.

brownag:
This is still causing problems this week. All of the functions referenced in this issue are still causing problems. 

After a bit of thought, I am going to create a new internal helper function that needs to be used to skip any tests or examples that rely on a fully-functioning terra installation.

The function will return a logical value: TRUE if we can initialize a SpatRaster and SpatVector object with a CRS properly, FALSE on warning or error.


Something like:
``` R
.terra_can_initialize_spatial_object <- function() {
  requireNamespace("terra", quietly = TRUE) && tryCatch({
    terra::rast(matrix(0), crs = "OGC:CRS84")
    terra::vect("POINT (0 0)", crs = "OGC:CRS84")
    # warning("smoke test")
    TRUE
  }, warning = FALSE, error = FALSE)
}
```

This can replace basically all usage of requireNamespace("terra") with the benefit of being a more complete check of whether the required functionality e.g. proj.db is there

PR #418, opened 2025-11-14
Broken proj

dylanbeaudette:
Addressing #375. Trying to find a reasonable comprise in function + tests that gracefully fail when sf/terra packages are missing proj.db (and maybe other, yet unknown problems).

### DIFF
--- a/R/ISSR800.R
+++ b/R/ISSR800.R
@@ -148,7 +148,8 @@ ISSR800.wcs <- function(aoi, var, res = 800, quiet = FALSE) {
     return(dl.try)
   }
   
-  # load pointer to file and return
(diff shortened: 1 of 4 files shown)


R/ISSR800.R
```diff
@@ -231,6 +232,9 @@ ISSR800.wcs <- function(aoi, var, res = 800, quiet = FALSE) {
       getOption('soilDB.return_Spatial', default = FALSE) && 
       requireNamespace("raster")) {
     r <- raster::raster(r)
+    
+    # return without setting metadata via terra methods
```
brownag: Should actually probably set metadata before converting to raster, rather than returning early

PR #420, opened 2025-11-17
feat: improve Soil Data Access metrics for soilDB usage

brownag:
This PR adds a more granular "SDA Query application" header comment to SDA queries following Phil Anzel's proposal for CART application. Low-level queries via `SDA_query()`, with no specific application context, already have the soilDB R package version included in the HTTP User-Agent header when POSTing queries.

To prevent having to put boilerplate handling into the many functions that call `SDA_query()`, I added simple logic to SDA_query itself to determine the calling environment. This will allow us to create comments that uniquely capture bare calls of SDA_query, user-defined functions, and functions defined in R packages. If we want to provide a great deal more specific info we may need to add custom handling, possibly via a new argument to pass query metadata to SDA_query.

TODO:
 - [x] Additional data elements (explicit list of: properties, interpretation names, aggregation methods? or simply include the raw function call?)
 - [x] Truncate to ensure comments are relatively short (much less than 3000 characters, and if we include the raw call explicitly avoid including e.g. a very large vector of mukeys) 
 
Handling of highest-level functions that run many queries (e.g. `fetchLDM()`, `fetchSDA()`) will result in several queries being generated each with their own comment. Some queries then may be associated with the top level fetch* function and some will be associated with lower-level get* or other internal functions. 

cc: @jneme910

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: soilDB
 Type: Package
 Title: Soil Database Interface
-Version: 2.8.13
+Version: 2.8.14
 Authors@R: c(person(given="Dylan", family="Beaudette", role = c("aut"), email = "dylan.beaudette@usda.gov", comment=c(ORCID="0009-0008-2780-4785")),
              person(given="Jay", family="Skovlin", role = c("aut")), 
              person(given="Stephen", family="Roecker", role = c("aut")), 


brownag:
I am sticking with the CART-like setup where the comment includes just application, rule, and version. 

e.g. `/** SDA Query application='soilDB' rule='soilDB::get_SDA_property' version='2.8.13' **/`

Rule is determined at runtime from the call stack, filtered to just soilDB. If there is some error in parsing call stack, or no calls to soilDB in the top 100 calls, then the `rule` safely falls back to `"soilDB::SDA_query"`.

At this time no other data elements will be included, so the two TODOs are resolved. We may revisit that depending on future metrics needs.

PR #434, opened 2026-01-09
add remarks argument to OSDquery

dylanbeaudette:
This PR adds a new `remarks` argument to `OSDquery()`. Searching the REMARKS section of an OSD can sometimes be useful for finding diagnostic horizons features. The related SoilWeb API has already been updated.

### DIFF
--- a/R/OSDquery.R
+++ b/R/OSDquery.R
@@ -23,7 +23,7 @@
 #' 
 #' @details 
 #' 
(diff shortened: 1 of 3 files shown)


R/OSDquery.R
brownag: Need to put `''san joaquin''` in code backticks (probably all of the syntax examples should be) because the default parsing converts these to double quotes, which does not work

c.f.
image

dylanbeaudette: Thanks, I missed that.

PR #435, opened 2026-01-11
adding geomorphon proportions by soil series to fetchOSD()

dylanbeaudette:
This PR makes available soil series -- geomorphon summaries (area-weighted proportions) available via `fetchOSD()`.

SoilWeb now maintains a snapshot of soil series -- geomorphon cross-tabulation, updated each SSURGO refresh cycle. Currently, this is only available for CONUS and at 30m resolution as defined by the current gSSURGO snapshot.

Geomorphons were derived from a 30m CONUS DEM, using GRASS GIS and the following:
```
r.geomorphon --o dem=elev30_int forms=forms30 search=75 skip=5 flat=1.718
```
Pixel-wise cross-tabulation at 30m resolution was performed using `r.stats` from GRASS GIS.

I plan to document these data and methods soon.

EDIT: This PR will also include a basic vignette outlining the purpose, background, and status of all functions and data sources which depend on SoilWeb. As @brownag pointed out, there are many details hidden behind functions like `fetchOSD()` that could use an explanation within soilDB.

### DIFF
--- a/vignettes/soilweb-data-functions.Rmd
+++ b/vignettes/soilweb-data-functions.Rmd
@@ -27,7 +27,7 @@ x <- fetchOSD(c('miami', 'pierre', 'tristan', 'lucy', 'musick'), extended = TRUE
 # Summary
 The soilDB package relies on a number of data sources curated as part of the [SoilWeb](https://casoilresource.lawr.ucdavis.edu/soilweb-apps) ecosystem of web applications, APIs, web coverage services (WCS), and internally-used lookup tables. These data are always updated shortly after the beginning of each fiscal year (October 1st) as part of the SSURGO refresh cycle. Within-cycle updates are typically performed quarterly. 
 
-Many of these curated data sources are documented in:
+Many of these curated data sources and their evolution through time have been documented in:
 
   * O'Geen, A., Walkinshaw, M. and Beaudette, D.E. (2017), SoilWeb: A Multifaceted Interface to Soil Survey Information. Soil Sci. Soc. Am. J., 81: 853-862. https://doi.org/10.2136/sssaj2016.11.0386n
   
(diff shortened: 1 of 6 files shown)


brownag:
There will be no CRAN related errors since you are using R_SOILDB_SKIP_LONG_EXAMPLES environment variable, but this also means the code will not be run except in environments where this env var is set to FALSE (only on our GH pages build)

Since it is a vignette, it will be visible on the CRAN site, and to users calling `vignette(package="soilDB")` without any of the code running or output generated. It is best to write the vignette with that in mind.

If you do not want it to be part of the package or visible on CRAN, we can add articles that are not vignettes to the GH site, but that is a different process and it is _not_ a vignette, strictly speaking. If that is what you want, I do not see the benefit of having it in the package repository, it would be fine to just link to custom-built HTML at https://ncss-tech.github.io/AQP/soilDB from the github site.

I also would prefer to not have igraph used for anything in this if it is a vignette. Though it appears to pass checks, if you are calling `requireNamespace("igraph")` in a vignette, igraph should be in suggests. And I do not want to add igraph to suggests. 

My recommendation in general is to focus on the data structures returned, and not on the visualization of them. The narrative should make sense with or without the code executed and knitted into the output.

vignettes/soilweb-data-functions.Rmd
```diff
@@ -0,0 +1,556 @@
...
+# series names listed in "competing" have the family classification as "series"
+knitr::kable(head(x$competing))
+
+# series names listed in "gas" are geographically associated with "series"
+knitr::kable(head(x$geog_assoc_soils))
+```
+
+
+One way of visualizing geographically associated soils, requires the igraph package.
+```{r, eval=igraphAvailable, echo=igraphAvailable, fig.cap="Geographically associated soils of selected soil series."}
```
brownag: Technically this `igraphAvailable` overrides the .Rmd wide R_SOILDB_SKIP_LONG_EXAMPLES logic. igraphAvailable probably should be a combination of requireNamespace and the environment variable, if it is going to be retained. I recommend removing these igraph-related sections and focusing on the data structures that soilDB returns.

brownag (commented):
Initial review comments

dylanbeaudette:
> There will be no CRAN related errors since you are using R_SOILDB_SKIP_LONG_EXAMPLES environment variable, but this also means the code will not be run except in environments where this env var is set to FALSE (only on our GH pages build)

Stepping back a moment. I agree that a vignette is probably the right type of document, however, fully documenting these data requires that `fetchOSD()` is able to run without error. It seems like most calls to web services in tests, examples, and vignettes are prone to triggering CRAN errors.

Any suggestions?

brownag:
> > There will be no CRAN related errors since you are using R_SOILDB_SKIP_LONG_EXAMPLES environment variable, but this also means the code will not be run except in environments where this env var is set to FALSE (only on our GH pages build)
> 
> Stepping back a moment. I agree that a vignette is probably the right type of document, however, fully documenting these data requires that `fetchOSD()` is able to run without error. It seems like most calls to web services in tests, examples, and vignettes are prone to triggering CRAN errors.
> 
> Any suggestions?

My suggestion is that you explain the result structure, without explaining specifics of the actual result of code being run. It will not cause issues with CRAN because you are setting eval=FALSE on CRAN.

vignettes/soilweb-data-functions.Rmd
dylanbeaudette: I'll consider moving the viz. stuff to linked articles. My intent was to highlight that these data are directed graphs.

PR #457, opened 2026-04-10
Update get_component_from_GDB.R

smroecker:
fixing case when WHERE is null


smroecker:
fixing when WHERE is null in get/fetch GDB

PR #477, opened 2026-06-17
perf(createSSURGO): improve reading and writing of tabular data

brownag:
This is a substantial performance improvement for reading and writing tabular data in `createSSURGO()`
 - use data.table::fread
 - add transactions and batching for writing tabular data
 - adds logging wrappers w/ `quiet=FALSE`


brownag:
I have a couple other features and general enhancements that these changes are alluding to (e.g. custom separators to support alternative flat file formats), but I will hold those off for a future PR. 

This should work with the original intended use case of createSSURGO run on output from downloadSSURGO or equivalent. I would also like to be able to support arbitrary CSV files following SSURGO schema, such as those dumped from a geodatabase, or derived from raster soil survey data source, or similar.

